PR #81, opened 2014-08-25
Bug fix: tmp dir not existing so radars.sqlite can't be downloaded.

asreimer:
Fixed an error with the temp dir not existing which prevents the radars sqlite file from being updated.

To test this bug fix, 
1) Delete your DAVIT_TMPDIR and then start davitpy. You will receive an error that says: 
"sqlInit() Error unable to open database file: /tmp/sd/.radars.sqlite"
This is a **big** problem because plotting and file reading depends on radars.sqlite being available.
2) From the command line, create your DAVIT_TMPDIR and then restart davitpy, this time you will see a message that says: "Radars information has been updated."
3) Delete your DAVIT_TMPDIR, install this bug fix, and then start davitpy. You won't receive the sqlInit() error.
4) Rejoice!

### DIFF
--- a/pydarn/radar/__init__.py
+++ b/pydarn/radar/__init__.py
@@ -37,8 +37,13 @@ except Exception as e:
 # Update local HDF5
 ####################################
 import os.path, time
-try:  dirn = os.environ['DAVIT_TMPDIR']
-except: dirn=os.environ['HOME']
+try:
+    dirn = os.environ['DAVIT_TMPDIR']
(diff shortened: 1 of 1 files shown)


loisks317:
So I recently installed davitpy and I have a problem with this when running ipython notebooks. I'm not sure where the temp directory is, and changing the lines in __init__.py in pydarns/radar didn't clear up the issue.

PR #109, opened 2014-10-01
implement logging instead of print statements

cmeeren:
This pull request replaces most print statements with the built-in logging functionality. This allows the user control over what is actually printed to the console: For example, running

```python
import logging
logging.getLogger().setLevel(logging.INFO)
```
will show INFO level messages and all more severe messages (WARNING, ERROR,  CRITICAL), while the ``logging.WARNING`` level will not show the INFO messages. By default, the WARNING level is active and thus INFO and DEBUG messages will not be shown.

Additionally, using the logging module allows for easy saving of logs to files, etc. The actual format of the logging messages can also be customized. All settings can be set both programatically and using config files (INI, JSON, YAML), but everything is left at defaults for now.

I have not touched any of the print statements where some kind of ``debug`` or ``verbose`` flag controlled their behaviour. I also did not touch print statements in ``if __name__ == '__main__'`` sections.

I also fixed two missing ``import sys``, which is not related to logging, but were so small and safe edits that I did them on the go.


cmeeren:
Now that the rc is implemented, what do we do with logging?

ksterne:
Can this be closed now that we've started implementing logging with the 0.5 version release?

asreimer:
Ya it can be closed.

PR #124, opened 2015-01-05
Feature: Adding plotAcf.py

asreimer:
Extends `pydarn/plotting` by adding the `plotAcf.py` file. This file provides two functions for plotting SuperDARN rawacf data, `plot_acf` and `plot_rli`. `plot_acf` visualizes the ACF/XCF for at an input range gate. `plot_rli` plots a range-lag-intensity plot, visualizing both the real and imaginary components of an ACF/XCF for an input `beamData` object.

To test this pull request, one simply needs to execute `python acfPlot.py`.

One may also test some of this code with the following:
    
    from datetime import datetime
    myPtr = pydarn.sdio.radDataOpen(datetime(2012,5,21),'sas', eTime=datetime(2012,5,21,2),fileType='rawacf')
    myBeam = myPtr.readRec()
    pydarn.plotting.acfPlot.plot_acf(myBeam,31)
    pydarn.plotting.acfPlot.plot_rli(myBeam)

The `__init__.py` file in the `pydarn/plotting` directory was modified to add an import all from `acfPlot.py`.

### DIFF
--- a/pydarn/plotting/__init__.py
+++ b/pydarn/plotting/__init__.py
@@ -12,12 +12,18 @@ This subpackage contains various plotting routines for DaViT-py
 	* :mod:`fan`
 	* :mod:`pygridPlot`
 	* :mod:`printRec`
+	* :mod:`acfPlot`
 """
 
 try:
 	from rti import *
 except Exception,e: 
 	print 'problem importing rti: ', e
+
+try:
+	from acfPlot import *
+except Exception,e: 
+	print 'problem importing acfPlot: ', e
 
 try:
 	from fan import *
--- a/pydarn/plotting/__init__.py
+++ b/pydarn/plotting/__init__.py
@@ -26,6 +26,7 @@ try:
 except Exception,e: 
 	print 'problem importing acfPlot: ', e
 
+try:
 	from iqPlot import *
 except Exception,e: 
 	print 'problem importing iqPlot: ', e
--- a/pydarn/plotting/acfPlot.py
+++ b/pydarn/plotting/acfPlot.py
@@ -1,57 +1,69 @@
+# Copyright (C) 2012  VT SuperDARN Lab
+# Full license can be found in LICENSE.txt
+# 
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# 
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-def calc_blanked(ltab,tp,tau,tfr,gate):
-  import numpy as np
+"""
+.. module:: acfPlot
+   :synopsis: A module for generating plotting ACF and XCF data
 
-  #Calculate the lags and the pulse table
-  lags = []
-  ptab = []
-  for pair in ltab:
-    lags.append(pair[1]-pair[0])
-    ptab.extend(pair)
-  lags=list(set(lags))
-  ptab=list(set(ptab))
-  lags.sort()  
-  ptab.sort()
+.. moduleauthor:: ASR, 20141230
 
-  txs_in_lag={}
-  for lag in lags:
-    txs_in_lag[lag]=[]
+*********************
+**Module**: pydarn.plotting.acfPlot
+*********************
+**Functions**:
+  * :func:`pydarn.plotting.acfPlot.plot_acf`
+  * :func:`pydarn.plotting.acfPlot.calc_blanked`
+  * :func:`pydarn.plotting.acfPlot.plot_rli`
+  * :func:`pydarn.plotting.acfPlot.nufft`
+"""
 
-  #number of ranges per tau
-  tp_in_tau = tau/tp
+def plot_acf(myBeam, gate, blanked=True, xcf=False, panel=0, ax=None):
 
-  #determine which range gates correspond to blanked samples
-  tx_times=[p*tp_in_tau for p in ptab]
-  blanked_samples=[]
-  for tx in tx_times:
-    blanked_samples.extend([tx, tx+1])
+    """Plot the ACF/XCF for a specified beamData object at a
+       specified range gate
 
-  for lag in lags:
-    for pair in ltab:
-      if (pair[1] - pair[0] == lag):
-        #which samples were used to generate the acf
-        S1=tp_in_tau*pair[0]+gate + 1.*tfr/(tp)
-        S2=tp_in_tau*pair[1]+gate + 1.*tfr/(tp)
-        br=[]
-        #check to see if the samples are blanked or not
-        if S1 in blanked_samples:
-          br.append(S1)
-        if S2 in blanked_samples:
-          br.append(S2)
-        txs_in_lag[lag].extend(br)
+    **Args**:
+        * **myBeam** : a beamData object from pydarn.sdio.radDataTypes
+        * **gate**: (int) The range gate to plot data for.
+        * **[blanked]**: (boolean) Specifies whether magnitude and 
+                           phase should be plotted instead of real and 
+                           imaginary.
+        * **[xcf]**: (boolean) Specifies wheather to plot XCF data or not
+        * **[panel]**: (int) from 0 to 3 specifies which data to plot of
+                       ACF/XCF, ACF/XCF amplitude, ACF/XCF phase, or
+                       power spectrum, respectively.
+        * **[ax]**: a matplotlib axis object
 
-  return txs_in_lag
+    **Returns**:
+        Nothing.
 
+    **Example**:
+        ::
+            from datetime import datetime
+            myPtr = pydarn.sdio.radDataOpen(datetime(2012,5,21), \
+                                          'kap',fileType='rawacf')
+            myBeam = myPtr.readRec()
+            pydarn.plotting.acfPlot.plot_acf(myBeam,24)
 
-
-def acfPlot(myBeam, gate, blanked=True, xcf=False, panel=0, ax=None):
+        Written by ASR 20141230
+    """
 
   from matplotlib import pyplot
   import numpy as np
   import pydarn
-  from matplotlib import colors
-  import matplotlib as mpl
-  import matplotlib.cm as cmx
                  
   number_lags = myBeam.prm.mplgs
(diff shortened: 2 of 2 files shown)


MuhammadVT:
The code is working fine on my machine (ubuntu 12.04), but is there a
reference to check the accuracy of the plots generated by the code? I took
the attached acf plot as reference (generated by a  ACF plotting and
fitting tool <http://superdarn/tiki-index.php?page=ACF+Plotting+Tools> on
our website), but couldn't compare it well with the test code results as
the scales are quite different.

Some Questions:

1) Considering someone who is not familiar with "beamData" object, isn't it
better to create "myBeam" inside the plot_acf function (as it is done in
fan.py, rti.py, ..., by feeding it with the necessary arguments like sTime,
eTime, rad, etc.) instead of giving "myBeam" as an argument by itself.
2) It might be more informative if related information like radar name,
date, beam number,..etc were added as title in the plot generated by
plot_rli. What do you think?
3) The test code is plotting for beam 5, but where is it specified in the
code?

asreimer:
Whoops, you should change the range gate in  `pydarn.plotting.acfPlot.plot_acf(myBeam,24)` to 31 and you'll see ground scatter that's blanked. I've edited my initial comment above. You'll notice that my plotting routine and the one on the website match except for a normalization constant. I have made the function plot in units normalized to the lag0 power. This should really be optional. I'll work on this soon.

I'll answer your questions in the order you asked:
1) My goal was to provide basic acf plotting functionality to davitpy. This of course can be improved upon. I find it more transparent and less "black box-y" to pass the myBeam object to these plotting routines. One could easily write a wrapping function that handles file fetching and plotting for a user as we currently do for rti plots, etc. The problem with those routines though is that they are all convolved together and not flexible at all (suppose I only want to plot the rti window from a specific beamData object to a pre-existing axis object). 
2) Agreed, I overlooked this, when I get a chance I'll add this.
3) This is not specified in the code, it is a function of what beam is loaded in the beamData object.

asreimer:
Forgot to mention that I appreciate your lightning fast response times @MuhammadVT!

aburrell:
Planning on checking this out on red hat/mac tomorrow.

asreimer:
Excellent! Looking forward to a productive year for davitpy development.

Happy New Year everyone!

MuhammadVT:
You are welcome asreimer :-)

aburrell:
Ok, I've tested it and have some suggestions:

1) improve the PEP8 compliance.  This would include (but not be limited to):
  A) Inconsistent use of whitespace (eg lines 328-332 where you use both x = y and x=y conventions)
  B) Remove unnecessary backslashes (eg lines 188-189)
  C) Remove whitespace before parenthesis and unnecessary parenthesis in conditional statements (eg if(x == y and a == b): instead of if  ((x == y) and (a == b)):)
  D) Lines span more than 72 characters (eg line 168)

It would also be nice if the comments consistently had a space after the pound sign, to make them easier to read.

2) While the code runs on both my mac and the older red hat machine, I can't test all of the output on the red hat machine because of the display settings.  This could be fixed by returning the figure handle from each of the plotting routines, or requiring that the user set up the figure and provides axis handles.

3) I'd recommend changing some of the labels to make them more publication-ready.  Suggestions are how I like to do things, probably other people won't like them so it's your call
  A) What are the units of Amplitude?
  B) suggest pwr_0 -> P_0
  C) suggest all units in LaTeX math font
  D)  (m/s) -> (m s^-1) (AGU desired unit format)
  E) Specify Phase units or make it obvious on the axis (use degree symbol or specify in fractions of pi)
  F) units on lag power
  G) Better units on Spectral Power (eg arb. units or arbitrary) 
  H) Suggest using suptitle instead of axis set_title on figure so it's clear that it's a whole-plot title instead of column titles.  Also, include titles on both plots if you don't go over to just providing the data by requiring axis handles.

To see what things look like:

test_default1_mac
test_default2_mac
test_panels_mac


test_panel1of2_redhat

asreimer:
@MuhammadVT, I have addressed all of your questions/comments with my response above and the above 2 commits.

MuhammadVT:
Ok Ashton.

by the way,  did you have a chance to address Angeline's list of comments?
If so we can merge this pull request.

aburrell:
We should test the changes before merging.

asreimer:
I still need to address some of @aburrell's suggestions before this is ready to be merged (as I get time). Feel free to run some testing in the mean time though.

asreimer:
Alright, I've been able to address @aburrell's suggestions now too. Here they are in order:

1) All PEP8 issues have been resolved. I'm using and IDE call pycharm and it's set up to complain if something doesn't follow PEP8; pretty snazzy.

2) Are the issues with plotting on Redhat specific to this pull request? Have you encountered similar issues with any of the existing plotting routines in davitpy? I haven't programmed these plotting functions any differently than those already part of davitpy. If it's an issue with your machines display settings, a better fix may be to change your matplotlib backend before executing any plot commands so that you plot to a file instead of on to the screen. You can even permanently change the settings in a matplotlibrc file.

3) In general, I'm hesitant to do too much hard coding of font sizes/styles/etc. since this stuff can be set up in a matplotlibrc file according to individual user preferences. Also, users can have full control over axes labels by building their own publication quality plots using the `ax` arguement. For example, it is unlikely that many people want to plot the power spectrum or both the complex ACF and the ACF power and ACF phase. That being said I agree with some of your suggestions and have implemented them:
A) Changed the label to "Lag Power" instead. The units are in Watts assuming a unit resistance.
B) `pwr_0` is the name of the parameter in the rawacf file. I'm hesitant to change this to `P_0` as it looks similar to `p_l` or `p_s` which are lag zero powers resulting from the data fitting process and are contained in fit files (`fitacf`, `fitex`, `lmfit`). Let me know what you think. Also, not sure why I had it as `r"pwr$_{0}$"` instead of `"pwr_0"`so I changed that too.
C) Not sure what you mean, could you provide me with an example? Thanks!
D) I would like to let users change this on a per publication basis as stated above.
E) This is a great idea. I've implemented it.
F) Units of power are are almost meaningless here. What is the resistance in the receiver? How much power is __actually__ received by a SuperDARN radar? We don't have a calibrated measure of the power on any radars from what I understand. Also, when normalized by the lag-zero power, the quantity is dimensionless.
G) Technically, for a power spectrum (or power spectral density) I supposed the units are really Watts/Hz, but again the units of power only really indicate that we are looking at the square magnitude of the voltage measured by the radar receiver. Probably best to remove units all-together. I've re-labelled it to read "Power Spectrum" Again, please let me know what you think!
H) Good idea! Didn't know I could do this... That's much better. Implemented this.

Whew, so there you have it. Test away and let me know if I need to change anything else.

aburrell:
Good evening @asreimer,

Pre-testing comments:

1) Huzzah!

2) It's specific to this routine, had no problem plotting fan, rti, etc on redhat.  Here's a code snippit I used in SpacePy that works well, no matter how a user chooses to set their display settings:

def plot_func(args, draw=True, figname=None):
    if draw:
        # Draw to screen.
        if plt.isinteractive():
            plt.draw() #In interactive mode, you just "draw".
        else:
            # W/o interactive mode, "show" stops the user from typing more 
            # at the terminal until plots are drawn.
            plt.show()

    # Save output file
    if figname is not None:
        plt.savefig(figname)

3) I agree with you wrt font size and style, but not with the unit format.  If you don't want to change it that's fine, so long as you don't think it will have much impact on people who want to use this routine.  I have yet to use any of the default plotting programs beyond testing, since I don't think they produce nice looking figures, so I'm hardly the target user.  That said, the changes you have made definitely make the resulting figures much clearer to me, and that's excellent.  Oh, and (C) was the most stylistic point, you won't want to implement it.  Basically using ($unit$) for units instead of (unit).

asreimer:
2) Hmm interesting. I'm going to have a thorough look at some of the other plotting routines to see if I can find any differences... it shouldn't behave like that.

3) I think that we should have a much broader/general conversation about cleaning up our plotting routines so that they can be presented in a consistent manner. Perhaps this is best accomplished through the dev mailing list, but maybe this is better accomplished through an online chat.

Btw thanks for clarifying on the units thing.

asreimer:
Only difference I can see that might make a difference is that `plotRti` and `plotFan` import `pylab` and my code does not. A quick test of this would be to try and use @w2naf's `musicPlot` and see if you can make figures (his code doesn't import `pylab` either).

aburrell:
It's not an issue of using/not using pylab.  The main difference is that the other two routines either hand back a figure handle, or have an option not to draw the plot to screen and allow the plot to be written to file.

MuhammadVT:
I have tested the code, the suggested points have been addressed as far as
I can notice.

When I run my vim-flake8, it gave me the following messages. But I don't
think the code violated PEP8, and the messages below can be neglected.
Correct me if I am wrong.

acfPlot.py|3 col 2| W291 trailing whitespace
acfPlot.py|8 col 2| W291 trailing whitespace
acfPlot.py|13 col 2| W291 trailing whitespace
acfPlot.py|45 col 72| W291 trailing whitespace
acfPlot.py|46 col 74| W291 trailing whitespace
acfPlot.py|48 col 74| W291 trailing whitespace
acfPlot.py|500 col 80| E501 line too long (116 > 79 characters)
acfPlot.py|517 col 80| E501 line too long (141 > 79 characters)
acfPlot.py|520 col 80| E501 line too long (89 > 79 characters)
acfPlot.py|535 col 1| W391 blank line at end of file

asreimer:
@aburrell Oh of course! I see now what you were saying. Thanks for your patience, sometimes I'm a bit dense... I've implemented the ability to plot to pngs or pdfs. We __really__ need to make a class that handles this.

@MuhammadVT your PEP8 checker is stricter than mine. Those trailing whitespace complaints were on inline comments and I've fixed them. The other complaints (lines 500 onward) are in the testing code. I'm not fixing that because we are going to eventually replace this code with proper unit tests.

asreimer:
The issues raised by @aburrell and @MuhammadVT have been addressed. Can I get someone to merge this in? Or am I cleared to merge it myself?

aburrell:
I’m happy for you to merge it.  Or we can have a test/merge party this afternoon?

asreimer:
I like the idea of a test/merge party. Let's do that.

aburrell:
@asreimer 

This looks good!  I have 1 request, for future readability, and that is to change the way that freq_scale_factor is defined:
acfPlot.py: Line 125

   freq_scale_factor = 1 / (lags[-1] * myBeam.prm.mpinc * 10.0 ** -6) \
        / (myBeam.prm.tfreq * 1000.) / 2. * (3. * 10 ** 8)

I recommend:
1) using 1. / (lags...) instead of 1 since everything is supposed to be floats
2) get rid of the "\" at the end of the line by using parenthesis
3) reformat the 1/(blah)/(blah)/(blah)x(stuff) and have (numerator) / (denominator)

asreimer:
Alright, I've improved the readability of that equation.

Issue #126, opened 2015-01-09
A bug in plotUtils.py


Shirling-VT:
When I ran the maps.ipynb in the notebook, the following error came out:
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-5-e40ba87ab2a6> in <module>()
      2 # Plot map
      3 subplot(121)
----> 4 m1 = plotUtils.mapObj(boundinglat=30., gridLabels=False)
      5 overlayRadar(m1, fontSize=8, all=True, markerSize=5)
      6 subplot(122)

/usr/local/lib/python2.7/dist-packages/davitpy-0.2-py2.7-linux-x86_64.egg/utils/plotUtils.pyc in __init__(self, datetime, coords, projection, resolution, dateTime, lat_0, lon_0, boundinglat, width, height, draw, fillContinents, fillOceans, fillLakes, coastLineWidth, coastLineColor, grid, gridLabels, showCoords, **kwargs)
    137 
    138     if draw:
--> 139       self.draw()
    140 
    141   def draw(self):

/usr/local/lib/python2.7/dist-packages/davitpy-0.2-py2.7-linux-x86_64.egg/utils/plotUtils.pyc in draw(self)
    183           merLabels = [False,False,False,False]
    184         # draw meridians
--> 185         out = self.drawmeridians(meridians, labels=merLabels, fmt=lonfmt,
    186                                  color='.6', zorder=10)
    187 

UnboundLocalError: local variable 'lonfmt' referenced before assignment

It turns out that the parameter "lonfmt" isn't assigned when gridLavels=False. I add two statements (indicated by "# added by shirling" ) to the function "def draw(self):" inside "plotUtils.py" to fix this bug, as is shown below:

        # label meridians on bottom and left
        if self.coords == "mlt":
          meridians = np.arange(0.,360.,15.)
          lonfmt = lambda x: "%02g"%(x*24./360.)                # added by shirling 
        else:
          meridians = np.arange(-180.,181.,20.)
          lonfmt = "%g"                                                         # added by shirling 
        if self._gridLabels: 
          if self.coords == "mlt":
            merLabels = [True, False, False, True]
          else:
            merLabels = [False,False,False,True]  
        else: 
          merLabels = [False,False,False,False]
        # draw meridians            
        out = self.drawmeridians(meridians, labels=merLabels, fmt=lonfmt,
                                 color='.6', zorder=10)

[...]

PR #127, opened 2015-01-11
Bugfix: lonfmt unassigned when gridLabels=False

asreimer:
This pull request fixes bug https://github.com/vtsuperdarn/davitpy/issues/126 reported by @Shirling-VT. Since I was touching this code, the pull request also includes changes to the `plotUtils.py` code for increased PEP8 compliance.

To test this bugfix, run through the `maps.ipynb` file using `davitpy-notebook`; you should no longer received the error encountered in https://github.com/vtsuperdarn/davitpy/issues/126.

Further testing can be done by running `python plotUtils.py` in the `utils` directory.

### DIFF
--- a/utils/plotUtils.py
+++ b/utils/plotUtils.py
@@ -12,242 +12,284 @@ Basic plotting tools
   * :func:`utils.plotUtils.mapObj`: Create empty map 
   * :func:`utils.plotUtils.genCmap`: generate a custom colormap
   * :func:`utils.plotUtils.drawCB`: draw a colorbar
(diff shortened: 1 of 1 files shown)


asreimer:
Normally you would of course test the pull request before merging it. Also, depending on the magnitude of the changes to the code, you might want to let a couple other people test it before merging too. When you start making your own pull requests, a general rule of thumb is to never merge your own pull request.

In this case, since the changes to the code were very small (besides the PEP8 edits) and the bug was obvious and easy to fix, it is probably alright for only one person to test and merge. Nice thing about git is that we can always revert if there is a problem.

PR #133, opened 2015-04-10
Enhancement rc params

w2naf:
Hi, everyone!

Ok, the SuperDARN meeting is coming up, and I would really like to see a new version of Master before we go.  Here is a pull request that I think is a step in the right direction.  It's really big, but here are the salient points:

* Ashton's enhancement_rc_params is used as a foundation.
* This does not actually create the rc_params functionality, but it does reorganize everything so you can just do "import davitpy"
* I've tested all of the notebooks and made fixes to make sure they all run with the re-organization.
* Notebooks use the new "%pylab inline" magic to be compatible with the recent iPython 3/Jupyter updates.
* I've made small changes to my MUSIC codes, mostly addressing plotting annotations.
* Ray Greenwald found some bugs in the Raytrace code, and those patches have been added.

Let me know what you think!  I would really like to see this get pushed to develop, then to master within the next few weeks.

All the best,
Nathaniel


w2naf:
Pull request made to master in error.  Please see #134 for pull request to develop.

PR #137, opened 2015-06-02
Develop: Back FoV

asreimer:
This implements @aburrell's code in pull request: https://github.com/vtsuperdarn/davitpy/pull/131. See that request for more details. 

All I've done is cherry picked the relevant commits in to a clean branch so that it could be merged more easily with the develop branch.

PR #139, opened 2015-06-24
Revert rti.py to version before pull request #137

asreimer:
Try out plotting an RTI with the develop branch and be amazed as your RAM usage goes way up in comparison to what it used to do (almost 4GB for 2 hours of data)... It also takes much longer to plot (more than 60x longer). And if you try to zoom in on anything, or move the plots around (interactive tools) it takes forever for anything to happen. FYI, I'm running a 4th gen i7 at 4GHz clock speed.

Something was severely broken when #137 was merged in so this pull request reverts rti.py back to the previous working version.

To test this out, you first need to be in the develop branch. Then you need to modify rti.py to add an `import numpy` call at the top of the code. Next, run this command:
`pydarn.plotting.rti.plotRti(datetime(2012,11,24),'sas', eTime=datetime(2012,11,24,2))`

Finally, switch to this pull request and run the same command (of course be sure to run `python setup.py install` after switching to a branch).

### DIFF
--- a/pydarn/plotting/rti.py
+++ b/pydarn/plotting/rti.py
@@ -32,863 +32,747 @@
   * :func:`pydarn.plotting.rti.drawAxes`
 """
 
-import datetime as dt
-import numpy as np
-import matplotlib.cm as cm
+
+import pydarn,numpy,math,matplotlib,calendar,datetime,utils,pylab
 import matplotlib.pyplot as plot
 import matplotlib.lines as lines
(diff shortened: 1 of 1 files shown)


asreimer:
Forgot to mention that this rollback does not break the backFov code. That all still works fine.

asreimer:
Due to the severity of this bug (can't really make rti plots...) I'm going to merge this in tomorrow morning unless someone yells at me and tells me not to.

PR #143, opened 2015-07-22
Hotfix for radInfoIo

ksterne:
Seems as though sys was left out of the hdwRead function here.  Oops.  Also, spellchecked.  The way the error messages are done has been cleaned up for the develop branch and so this doesn't apply there.


asreimer:
Looks good to me. Merging.

PR #144, opened 2015-07-28
Added script for Fedora 21 install.

bsinglet:
Hey @asreimer,

Here's the commit for my Fedora install script.  Since yum has been deprecated, I only added support for dnf, although the syntax and package names are the same for both programs.  
Packages named differently than on Ubuntu:
-gcc-fortran
-hdf5-devel
-scipy
-basemap (needed to experiment with the flags to get this one to work)

Packages added:
-geos-python
-geos-devel
-Cython (this used to be installed automatically with h5py, but that breaks the install process in many instances, including mine.  Installing it explicitly first works better.)

I need to test on a clean install to make sure there aren't other dependencies that are needed that I happened to have installed, but otherwise these seem to be the only ones that are different from the Ubuntu install.

### DIFF
--- a/install/fedora_dependencies.sh
+++ b/install/fedora_dependencies.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# Python install script for Fedora
+#	installs all pre-requisite software to run DaViT-py
+#	tested on Fedora 21
+
+ver=2.7
+
+dnf install -y python$ver
+dnf install -d python-dev
+dnf install -y python-pip
+dnf install -y python-zmq
+dnf install -y python-imaging
+dnf install -y mpich2
+dnf install -y gcc-gfortran
+dnf install -y hdf5-devel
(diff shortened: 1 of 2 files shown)


asreimer:
This is great @bsinglet, thank you!

When I'm back home (away until Aug 7th), I'll fire up my Fedora VM and give this a test.

asreimer:
@bsinglet, we just made a new pre-release of davitpy, release 0.4-master. Can you update this pull request (pull down the upstream master)? Also, I've made a pull request on your branch (https://github.com/bsinglet/davitpy/pull/1) to fix a couple minor issues that prevented me from installing davitpy for Fedora 22.

asreimer:
To speed this up so we can have it merged right away, I have created a pull request (https://github.com/vtsuperdarn/davitpy/pull/177) with @bsinglet's original commit and my edits to fix bugs. Closing this one.

Issue #120, opened 2014-11-06
Problem with Rti mag plotting


guohucn:
Hey, guys. I am new here, and not familiar with davitpy.
I'v got a problem as follows:
input:
  t1=dt.datetime(2013,3,17,11,50)
  t2=dt.datetime(2013,3,17,14,0)
  fig=pydarn.plotting.rti.plotRti(t1,'cvw',t2,bmnum=4,params=['velocity'],coords='mag')
output:

![figure_1](https://cloud.githubusercontent.com/assets/9592043/4938007/7bd09082-65c8-11e4-8664-0ecb8716d429.png)

It looks like the data above 52 degrees are cut off. There should be a SAPS.
Everything is OK if I use geo as coords. 
![figure_2](https://cloud.githubusercontent.com/assets/9592043/4938644/ece7fb1c-65cc-11e4-86e3-8bb350e2b873.png)
I use the develop branch. The master branch is even worse.  
By the way. The month in the title and the time along X-axis are incorrect. This is also a minor problem. 
Thanks!

asreimer:
This is now fixed in #149. The problem was the beam 4 of CVE increases and then decreases in magnetic latitude. This means that when plotting as a function of mlat, the far range gates without data in them were getting plotted as white squares on top of the data at nearer range gates.

I fixed this bug by masking range gates without data in them and then using the `set_bad` method on the colormap to set the alpha to 0 for the "bad" (masked) range gates. This allows us to still use the speed of pcolormesh for plotting.

MuhammadVT:
Hi Ashton,

I tested that the bug in #120
<https://github.com/vtsuperdarn/davitpy/issues/120>has been fixed by this
pull request.

A very minor issue is that it looks like the function arguments in the doc
strings are not matching the corresponding functions' arguments quite well,
and the examples are not updated, for example, the doc strings of
plot_skynoise, plot_searchnoise, plot_freq, etc.  Would you mind going
through the doc strings one more time?

Also, the argument fmt in line 750
<https://github.com/vtsuperdarn/davitpy/pull/149/files#diff-20f7739eb5ff53dad0b5ab2d74817b01R750>
of the plot_nave function should be  'k:'  instead of  'k-'  , right?


Thanks

<https://github.com/vtsuperdarn/davitpy/issues/120>

ksterne:
I agree here with @MuhammadVT , it seems as though this is fixed in #149.

PR #149, opened 2015-09-18
RTI Plotting Refactoring

asreimer:
This pull request fixes a few issues with the RTI plotting and also refactors the code into a more modular form. The refactoring allows one to make custom RTI plots much more easily than could be done in the past. 

One major issue that has been fixed is the x-axis labeling. Time is now properly displayed and formatted according to whether the data crosses changes in date or multiple days.

Some variable names have been updated, but to prevent this from breaking anyone's existing code I've added a deprecation warning and a temporary wrapper. Specifically, `plotRti` is being deprecated in favour of `plot_rti`.

To test this code run the following 3 commands and compare your figures with mine:

1) `f = pydarn.plotting.plotRti(datetime(2012,11,26,15),'kod',eTime=datetime(2012,11,27,2),channel='all',params=['velocity','velocity_error'],coords='mag')`

you should see a deprecation warning too!

2) `f = pydarn.plotting.plot_rti(datetime(2012,11,25),'kod',eTime=datetime(2012,11,27),channel='all',params=['velocity','velocity_error'],coords='gate')`

3) `f = pydarn.plotting.plotRti(datetime(2012,11,25),'kod',eTime=datetime(2012,11,25,15),channel='all',params=['power'],coords='geo')`

1)
figure_1
2) 
figure_2
3)
figure_3

### DIFF
--- a/davitpy/pydarn/plotting/rti.py
+++ b/davitpy/pydarn/plotting/rti.py
@@ -43,12 +43,29 @@ from davitpy.utils.timeUtils import *
 from davitpy.pydarn.sdio import *
 from matplotlib.figure import Figure
 
+def plotRti(sTime,rad,eTime=None,bmnum=7,fileType='fitex',params=['velocity','power','width'],
+            scales=[],channel=None,coords='gate',colors='lasse',yrng=-1,gsct=False,lowGray=False,
+            show=True,filtered=False,fileName=None, tFreqBands=[],myFile=None,xtick_size=9,
+            ytick_size=9,xticks=None,axvlines=None,plotTerminator=False):
+    """
+    Wrapper for plot_rti. This function is being deprecated.
+    """
+    print "Warning: This function is being deprecated. Use plot_rti instead."
+    print "Calling plot_rti."
 
-def plotRti(sTime,rad,eTime=None,bmnum=7,fileType='fitex',params=['velocity','power','width'], \
-              scales=[],channel=None,coords='gate',colors='lasse',yrng=-1,gsct=False,lowGray=False, \
-              pdf=False,png=False,dpi=500,show=True,retfig=False,filtered=False,fileName=None, \
-              custType='fitex', tFreqBands=[], myFile=None,figure=None,xtick_size=9,ytick_size=9,xticks=None,axvlines=None,plotTerminator=False):
-  """create an rti plot for a secified radar and time period
+    return plot_rti(sTime,rad,eTime=eTime,bmnum=bmnum,fileType=fileType,params=params,scales=scales,
+                    channel=channel,coords=coords,colors=colors,yrng=yrng,gsct=gsct,low_gray=lowGray,
+                    show=show,filtered=filtered,fileName=fileName,tfreqbands=tFreqBands,myFile=myFile,
+                    xtick_size=xtick_size,ytick_size=ytick_size,xticks=xticks,axvlines=axvlines,
+                    plot_terminator=plotTerminator)
+
+def plot_rti(sTime,rad,eTime=None,bmnum=7,fileType='fitex',params=['velocity','power','width'],
+             scales=[],channel=None,coords='gate',colors='lasse',yrng=-1,gsct=False,low_gray=False,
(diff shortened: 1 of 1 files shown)


asreimer:
This pull request *could* be used to fix bug https://github.com/vtsuperdarn/davitpy/issues/120. I can do this tomorrow.

MuhammadVT:
Hi

I have tested the three commands for this pull request and got the expected
results. But the bug in #120
<https://github.com/vtsuperdarn/davitpy/issues/120> is still not been fixed
by this pull request yet. Also, it would be nicer if the y-axis ticks for
latitudes are located at integer points instead of floating points.

Thanks

MuhammadVT:
Hi again,

I have just found a bug in the cold. if you run
f = pydarn.plotting.plotRti(datetime(2012, 11, 26, 15), 'kod',
eTime=datetime(2012, 11, 27, 2), channel='all', params=['velocity'],
coords='mag', scales=[-400, 400])
it gives error message. It is because when "params" is set to a single
value like ['velocity'], len(params) = 1 but len(scales) = 2, which
violates the assertion in line 128

However, running it in the following way fix the problem
f = pydarn.plotting.plotRti(datetime(2012, 11, 26, 15), 'kod',
eTime=datetime(2012, 11, 27, 2), channel='all', params=['velocity'],
coords='mag', scales=[[-400, 400]])

So, the easiest way to fix the bug might be to make the "scales" argument
as nested list even when "params" is set to a single parameter.

Thanks

asreimer:
Bug #120 is now fixed with my latest commit. Here's the figure that I get with davitpy:

figure_1

and here's what you get with the VT website:

rti

The only difference is the coordinates are a bit different. I attribute this to the fact that AACGM in DaViTpy has been patched to fix an error with the wrong coefficients file being loaded. I suspect the VT website is using AACGM without that fix.

asreimer:
Hey @MuhammadVT 

The bug you found is not a bug, however, the asserts could be clearer. I've added an assert that makes sure that scales is input they way it needs to be, but provides a clear error message if it isn't.

I have modified the formatting for the yaxis on the RTI panels. Should be better now!

MuhammadVT:
Hi @asreimer

I came across a few minor issues in the refactored rti.py while I was about
to merge this pull request. Here are the list:

1) in line 749, the comment does not match with content. Shouldn't it be *#plot
the nave data*  ?
2) in line 750, it should be *fmt='k:'* instead of
*fmt='k-'*
3) The **Args** in some functions do not match with the corresponding
functions' actual arguments. Would you mind going over all the ""Args"" and
delete extra arguments?

After this, we should merge this to the new release.

Thanks

ksterne:
Hey @asreimer,

You'll see the several commits I added to PEP8 style the rti.py file.  I also made the simple edit that @MuhammadVT noted on the number of averages.  Maybe just a copy and paste fail there.  I re-ran through all of the 4 test (the 3 you mentioned and the fix for #120 ) and it looks like they are check out.  Do you want to make sure I didn't mess anything up with the PEP8 styling?  I'll wait to merge after one last look through.

asreimer:
Hey @ksterne and @MuhammadVT. I have checked all the code after the PEP-8 edits and everything still works. I have also cleaned up the documentation as requested. 

Run through the tests in my original post and see if everything is working for you.

MuhammadVT:
I have run through the following commands, and the results are as expected.

1) f =
pydarn.plotting.plotRti(datetime(2012,11,26,15),'kod',eTime=datetime(2012,11,27,2),channel='all',params=['velocity','velocity_error'],coords='mag')

2) f =
pydarn.plotting.plot_rti(datetime(2012,11,25),'kod',eTime=datetime(2012,11,27),channel='all',params=['velocity','velocity_error'],coords='gate')

3) f =
pydarn.plotting.plotRti(datetime(2012,11,25),'kod',eTime=datetime(2012,11,25,15),channel='all',params=['power'],coords='geo')

4) fig=pydarn.plotting.rti.plotRti(datetime(2013,3,17,11,50),'cvw',datetime(2013,3,17,14,0),bmnum=4,params=['velocity'],coords='mag')

We can go ahead and merge this.


Thanks

PR #152, opened 2015-09-22
Bugfixes printRec

ksterne:
Found a number of little bugs when trying to use the develop version of printRec today.  The biggest change here is that the beamData.channel value is an integer and not a string as it previously was.  Treating channel as a string was breaking the code is several places.  Made sure to update this in the radDataTypes documentation.  Another big thing is the radDataOpen call had not been updated to use the new formats...so the UAF.channel filenames would not be found when calling this function. Added a little bit of commenting to printRec as well.

To test, below some sample code.  I encourage you to try out radar for bks, pyk, ade, mcm in combination with changing the summ=0 value to either a 0 or 1.  summ=1 will just output a beam summary for each sounding and not all of the range gates with scatter.  This code will output the results in the out.txt file.

#!/usr/bin/env python

import os
import tempfile
os.environ['MPLCONFIGDIR'] = tempfile.mkdtemp()
import matplotlib
from davitpy import pydarn
import datetime as dt

year = 2015
month  = 8
day = 20
shr    = 0
smin = 30
ehr = 1
emin = 0
radar   = 'pyk'
filename = 'out.txt'
filetype = 'fitacf'

pydarn.plotting.printRec.fitPrintRec(dt.datetime(year,month,day,shr,smin),dt.datetime(year,month,day,ehr,emin),radar,filename,summ=0,fileType=filetype)

### DIFF
--- a/davitpy/pydarn/plotting/printRec.py
+++ b/davitpy/pydarn/plotting/printRec.py
@@ -170,6 +170,8 @@ def fitPrintRec(sTime, eTime, rad, outfile, fileType='fitex', summ=0):
   while(myData is not None and myData.time <= eTime):
     t = myData.time
     if(summ == 0):
+      # If not interested in the summary, lets print all of the range gate values.  We'll
+      # start with the header for each beam sounding.
       f.write(t.strftime("%Y-%m-%d  "))
       f.write(t.strftime("%H:%M:%S  "))
       f.write(radar.name+' ')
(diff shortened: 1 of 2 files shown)


ksterne:
I should also note that I added to the end a close to the pointer that's opened here.  It might not be necessary, but maybe helps clean things up.

MuhammadVT:
Hi,

I have tested it for bks, pyk, ade, mcm while changing the "summ" value to
either 0 or 1. The bug is fixed by this pull request.

Thanks

Issue #119, opened 2014-10-29
Iterative data reading or memory leak


ksterne:
Moving this off of the google group and into github since it seemed as though it was a problem with the code and not something that I was doing.  

Again, here is a summary of the problem I'm seeing:

I'm running some code that reads in rawacf files for an entire day for one radar, searches for certain characteristics and then moves on to the next day in repeat. I'm doing a radDataOpen for each day.  However, after about a month and half or maybe two months of doing this, the code crashes as something seems to fall apart in the copying of data to the local machine as well as the concatenation.

The very basic of what I'm trying to run can be found in my github repository at:
https://raw.githubusercontent.com/ksterne/radops/master/acfd_find.py

ksterne:
This was bad programming on my part as you see from the example code that I open the myPtr, but never close it.  Added a myPtr.close() to the end and I think it's working much better.

asreimer:
Good to know! Thanks!

PR #153, opened 2015-09-24
Closing the opened sftp connection

ksterne:
I think this was related to some issues I was finding when I brought up #119 .  One of the issues was the local computer running out of memory, but that was just bad programming on my part.  Garima and I have off and on been running iterative type code which is pulling over lots of data in one go.  I believe that if you're iterating through data open calls (and even if you close the pointer) the sftp connection stays open to the sftp server.  For me this popped up as I would run my code and then a few hours later see that the sd-data1.ece.vt.edu computer had over 1000 processes on it, most of which were sshd processes for sd_dbread.  

I think this might not be an issue for most people as I think the sftp connection is closed whenever the initial process/code that started everything is done.  However, for iterative type data call and processing, it starts to be a problem.  In general, it's probably good that we close the sftp connection once we've downloaded all of the files that are needed.

I don't have a good way for people to test this that are outside of VT and know how to see the processes running on the sd-data1.ece.vt.edu sftp server.  You would have to setup your own sftp server, have some code loop through a bunch of radDataOpen calls (a start on this type of code could be pulled from #119), and then watch the number of sshd processes increase on the sftp server while the code is still running.  Flip it over to this branch and the sshd processes will go away once the files have been downloaded.  

Otherwise, you could coordinate something with me over e-mail/google-hangouts and I can watch the processes increase here while you run some code.

### DIFF
--- a/davitpy/pydarn/sdio/fetchUtils.py
+++ b/davitpy/pydarn/sdio/fetchUtils.py
@@ -699,6 +699,11 @@ def fetch_remote_files(stime, etime, method, remotesite, remotedirfmt,
     # after deleting the dictionary structure containing the password
     del remoteaccess
 
+    #--------------------------------------------------------------------------
+    # Close the opened sftp connection
+    sftp.close()
+    transport.close()
+
     return filelist
 
 
--- a/davitpy/pydarn/sdio/fetchUtils.py
+++ b/davitpy/pydarn/sdio/fetchUtils.py
@@ -701,8 +701,9 @@ def fetch_remote_files(stime, etime, method, remotesite, remotedirfmt,
 
     #--------------------------------------------------------------------------
     # Close the opened sftp connection
-    sftp.close()
-    transport.close()
+    if method is "sftp":
+	    sftp.close()
+	    transport.close()
 
     return filelist
 


asreimer:
We haven't been good about these kinds of things in our code. Nice catch!

I won't be able to test this today, but I can test this against our server at usask.

ksterne:
Does anyone know if we also need to close the urllib connections if those are used?  Maybe @aburrell could look at something like that or anyone else that has a non-sftp type server they can pull data from?  I was mostly focusing on the sftp connection here since I've seen the sshd processes increase as I keep iteratively opening up a month's worth of data.

ksterne:
Hey @asreimer, think you'll get around to testing this?  I'll need to pipe this into the release branch if this gets through.

ksterne:
Bump.  If no one else wants to give this a look, can it be merged maybe?  @asreimer, any hope on getting you to look at it with the setup there?  I'll need to hotfix this into release as well.

aburrell:
Sorry, I can look at this next week.  I’ve solved this before but had trouble committing.  I think I’ve got github set up better now so it shouldn’t be a problem.

ksterne:
Bump.  Anyone?

PR #159, opened 2015-10-14
Tepidfix raytrace / 2 Parameter Method

w2naf:
* Sets Raytrace ranges out to 4000 km by setting an 8 km horizontal step size.
* Adds 2 notebooks illustrating the 2 Parameter Method Geolocation algorithm [Greenwald et al., 2015, Radio Sci., Submitted]


salsrag:
Nathaniel,

 

There is one issue in your commit summary. This version does not work on Southern-Hemisphere radars. You have to be careful with the magnetic field geometry in the Southern Hemisphere. I played with it a bit, but wasn’t sure it was working correctly, so I set it aside. 

 

Ray

w2naf:
Does this apply to your 2 Parameter Method Notebooks only? Or does this also apply to the raytrace code itself?

Nathaniel

salsrag:
I am pretty sure it applies to the two parameter analysis only. I will look at the code again to remember wha the exact issues are.

 

Ray

asreimer:
Closing this since it is essentially a duplicate of #160 now that the 0.4-master release has gone out.

PR #163, opened 2015-10-16
Release1.0 readfit

garimamalhotra:
After this, we will be able to read 'fit' files like we read 'fitacf' files. This routine converts a 'fit' to 'fitacf' while opening only, so that rest of the functions keep behaving in the same way as before.

from davitpy.pydarn.sdio import radDataRead
import datetime as dt

myptr=radDataRead.radDataOpen(dt.datetime(2002,8,19,1,0,0,0),'sas',dt.datetime(2002,8,19,2,0,0,0),\
channel=None,fileType='fit', filtered=False,src='sftp')

"""
Try with all these commands:
myptr=radDataRead.radDataOpen(dt.datetime(2009,8,19,1,0,0,0),'sas',dt.datetime(2009,8,19,2,0,0,0),\
channel=None,fileType='fitacf', filtered=False,src='sftp')

myptr=radDataRead.radDataOpen(dt.datetime(2002,8,19,1,0,0,0),'sas',dt.datetime(2002,8,19,2,0,0,0),\
channel=None,fileType='fit', filtered=False,src='local')

myptr=radDataRead.radDataOpen(dt.datetime(2009,8,19,1,0,0,0),'sas',dt.datetime(2009,8,19,2,0,0,0),\
channel=None,fileType='fitacf', filtered=False,src='local')

"""
x=1
mybiglist=[]
while(x!=None):
        x=radDataRead.radDataReadScan(myptr)
        mybiglist.append(x)

### DIFF
--- a/davitpy/pydarn/sdio/radDataTypes.py
+++ b/davitpy/pydarn/sdio/radDataTypes.py
@@ -171,9 +171,9 @@ class radDataPtr():
             else:
                 os.system('cp '+fileName+' '+outname)
                 print 'cp '+fileName+' '+outname
-            if fileType=='fit':    #If a fit file name is given, it is converted to fitacf
+            if fileType=='fit':    #If a fit file name is given, it is converted to fitacf, as our dmap routines can't read fit files
                 outname_new="_".join((outname,'fitacf'))
-                print "Converting fit to fitacf"
+                print "Can't read fit files. Converting to fitacf.."
                 os.system('fittofitacf '+outname+' > '+outname_new)
                 outname=outname_new
             filelist.append(outname)
@@ -264,6 +264,7 @@ class radDataPtr():
                                              local_fnamefmt, verbose=verbose)
 
                 if ftype=='fit':   #If fit files found, converting them to fitacf files
+                    print "Can't read fit files. Converting to fitacf.."
                     temp_new=[]
                     for f in temp:
                         new_name=string.replace(f,'fit','fitacf')
@@ -367,6 +368,7 @@ class radDataPtr():
                     password=password, port=port, verbose=verbose)
 
                 if ftype=='fit':   #If fit files found, converting them to fitacf files
+                    print "Can't read fit files. Converting to fitacf.."
                     temp_new=[]
                     for f in temp:
                         new_name=string.replace(f,'fit','fitacf')
--- a/davitpy/pydarn/sdio/radDataTypes.py
+++ b/davitpy/pydarn/sdio/radDataTypes.py
@@ -238,9 +238,7 @@ class radDataPtr():
                     local_dict = {'radar':radcode, 'ftype':ftype, 'channel':channel}
                 if ('ftype' in local_dict.keys()):
                     local_dict['ftype'] = ftype
-                #import ipdb
-                #ipdb.set_trace()
-                if ftype=='fit':
+                if local_fnamefmt is None or ftype=='fit':
                     local_fnamefmt=['{date}{hour}.*{ftype}'] #File name format for fit files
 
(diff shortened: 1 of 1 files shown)


MuhammadVT:
Hi Garima,

I have tested this pull request, and it is working fine. Would you mind
adding a print statement after the  *fittofitacf *routine to explain what
is happening there? it might confuse an outside user who is not aware why
we do fittofitacf conversion.

Thank you.

garimamalhotra:
Hey Muhammad,

I have added print statements. Let me know, if this is what you wanted.

Thanks

asreimer:
Hey @garimamalhotra,

This is a nice little fix, but it uses the RST binary `fittofitacf` which requires a user to have RST or RSTLite installed on their machine otherwise they won't have the `fittofitacf` program. In my opinion, it would be best if the user doesn't have to install RST/RSTLite (in fact, we worked to remove the dependency on RST/RSTLite from DaViTpy some time ago, we still have the dmap stuff in C but @jspaleta has a purely python version that we could use instead).

From a software development and codebase maintainability standpoint, it would be much better if instead did either:

1) added code that enabled DaViTpy to read .fit files natively
2) had the SuperDARN data distribution people provide a repository of fitacf files that have been converted from the fit files that are kicking around.

By bringing RST/RSTLite dependence back in to DaViTpy we are moving away instead of towards the ability to install and use DaViTpy on windows machines. It also makes the installation process in general more difficult for any user.

garimamalhotra:
I agree to that !

MuhammadVT:
Hey @Ashton <https://github.com/asreimer>

Good to hear back from you.

Garima is doing this as per Mike's request, just to work around the problem
of not being able to read fit files in davitpy. The Conversion from fit to
fitacf is because the routine in RST for reading fit files directly is not
working. As you said, the best way is to read the fit files directly.
Actually, we are also going to convert all the fit files to fitacf files
very soon.

asreimer:
Hey @MuhammadVT,

That's fine. I don't really have a problem with this being merged in as a temporary fix so long as there is a proper fix being worked on. While adding `fittofitacf` code into davitpy fixes the problem temporarily, it re-introduces the requirement that people have RST/RSTLite installed on their machine. This is not a good thing so I don't think we should be doing that.

Instead of hard-coding a solution straight into davitpy like this, why not do something like this:
1) fetch required fit file using `fetchUtils`
2) convert fit file to fitacf file 
3) use davitpy to read/plot/whatever with fitacf file

The above has the advantage of fixing your local VT problem without bringing RST/RSTLite in as a dependency for davitpy. An added benefit is that we don't have to strip this out of the davitpy codebase once you guys have converted all your fit files to fitacf.

I'm a bit confused by what you said though, "The Conversion from fit to fitacf is because the routine in RST for reading fit files directly is not working." Which RST routine is not working? Do you mean that davitpy isn't able to read fit files directly? 

AFAIK .fit files and .fitacf files are of a completely different file format. .fitacf files are a `dmap`

MuhammadVT:
I agree with you. I thought fittofitacf code is in davitpy already.

Evan was doing some work which requires reading fit file directly using a
routine in RST in our davit code base.  I was told that reading fit file
breaks in somewhere. I will ask him about the routine he has used and will
let you know.

Yes, fit ant fitacf files are in different formats.

ksterne:
Should this pull request be changed to go into develop instead of the now defunct release branch?

garimamalhotra:
I can work on a better solution for this as @asreimer suggested .

asreimer:
Hey @garimamalhotra, I opened issue #174 to discuss solutions to this problem.

PR #164, opened 2015-10-19
Bugfix close sftp

ksterne:
Adding this hotfix request to the release branch as it seems as though this should be corrected before we make a new master branch on the 21st.  Will note this hasn't been merged in to the develop branch.  Does have a few pending testers I believe though.


MuhammadVT:
Hi @Kevin <https://github.com/ksterne>,

Is this the same as #153 <https://github.com/vtsuperdarn/davitpy/pull/153>.
I will test it by tomorrow morning. But it would be great if @aburrel
<https://github.com/aburrell> could take a look as she is the Author of
fetchUtils.py

aburrell:
I tested this and it does the job.

PR #170, opened 2015-10-29
Pull release of current master (0.3) into master-old

asreimer:
This is a pull request to make the current release (0.3) the old master in the wake of release-0.4.


ksterne:
Seconding this notion as a way to give people a backup for any reason.

PR #172, opened 2015-10-30
Displaying plots

aburrell:
Added option to display or repress plots.


aburrell:
Changes made in response to @asreimer 's comments while testing.

PR #177, opened 2015-11-03
Addition of Dependency Install Script for Fedora

asreimer:
This pull request adds a dependency install script for Fedora 21 and 22. The original work on building this script was performed by @bsinglet and I have included some bug fixes that enable raydarn dependencies to be installed. Also removed old code that added stuff to .bashrc since we don't need that anymore (environment variables aren't used as of 0.4-master release).

To test this, you need to start with a fresh install of Fedora 21 or 22 and then run this script.

### DIFF
--- a/install/fedora_dependencies.sh
+++ b/install/fedora_dependencies.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# Python install script for Fedora
+#	installs all pre-requisite software to run DaViT-py
+#	tested on Fedora 21
+
+ver=2.7
+
+dnf install -y python$ver
+dnf install -d python-dev
+dnf install -y python-pip
+dnf install -y python-zmq
+dnf install -y python-imaging
+dnf install -y mpich2
+dnf install -y gcc-gfortran
+dnf install -y hdf5-devel
+dnf install -y python-matplotlib
+pip install --upgrade matplotlib
+dnf install -y python-basemap-data
+pip install --upgrade ipython
+dnf install -y ipython-notebook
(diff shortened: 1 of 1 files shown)


asreimer:
I've tested this on Fedora 23 and it works there too.

MuhammadVT:
I am working on this. Will merge it by tomorrow if it works fine.

MuhammadVT:
I have tested this pull request on clean installations of Fedora 21, 22, 23 in VM VirtualBox. In all cases I am getting "IOError: [Errno 30] Read-only file system" after line 17 (dnf install -y gcc-c++) and a bunch of "Bus errors" starting from line 20 (pip install --upgrade matplotlib). Any idea why this is happening?

It installs python3.4 under /usr/lib64 as well when testing on Fedora 23. So we may need to specify python version number explicitly.

asreimer:
That IOError is weird... No idea why you are getting that.

As for the python3.4 thing, we might have to do something to specify the python version. Thanks for testing this, when I'm home from AGU I'll setup a fresh Fedora VM and try to reproduce your issues. My previous Fedora 23 test was not on a fresh system install (in hindsight, not surprisingly important).

MuhammadVT:
I found that the problem I have faced in testing this pull request is due to way I installed the Fedora.  The testing failed when I chose "Try Fedora" option to install Fedora to my Virtual Box, which gave the above mentioned error messages. When I tried "Install to Hard Drive" option, I could be able to test this pull request successfully on all of Fedora 21, 22, 23. So I am merging this pull request.

PR #185, opened 2015-11-13
BugFix: plot_rti fails if no elevation angle data available.

asreimer:
As the title suggests, I've found a bug in plot_rti that is caused by no elevation angle data being present in the data file. 

To test this pull request, first run the following code on the current develop branch:

    from davitpy.pydarn.plotting import plot_rti
    from datetime import datetime,timedelta
    stime = datetime(2014,7,10)
    etime = stime+timedelta(days=1)
    
    fig = plot_rti(stime,'sas',params=['elevation'],bmnum=2,eTime=etime,fileType='fitacf')

You will get an error message like this:

>/home/reimer/.local/lib/python2.7/site-packages/davitpy-0.2-py2.7-linux-x86_64.egg/davitpy/pydarn/plotting/rti.pyc in rti_panel(ax, data_dict, pArr, fplot, gsct, rad, bmnum, coords, cmap, norm, plot_terminator)
>   1061             for j in range(len(data_dict['slist'][fplot][i])):
>   1062                 if(not gsct or data_dict['gsflg'][fplot][i][j] == 0):
>-> 1063                     data[tcnt][data_dict['slist'][fplot][i][j]] = pArr[i][j]
>   1064                 elif gsct and data_dict['gsflg'][fplot][i][j] == 1:
>   1065                     data[tcnt][data_dict['slist'][fplot][i][j]] = -100000.
>
>TypeError: 'NoneType' object has no attribute '`__getitem__`'

This is caused by a radar control program not having any interferometer data and then the pArr[i] list is of NoneType.

This pull request fixes that problem by detecting when pArr[i] is None to simply skip over it in the loop.

### DIFF
--- a/davitpy/pydarn/plotting/rti.py
+++ b/davitpy/pydarn/plotting/rti.py
@@ -175,7 +175,7 @@ def plot_rti(sTime, rad, eTime=None, bmnum=7, fileType='fitex',
     assert(isinstance(eTime, datetime.datetime) or
            eTime is None), 'error, eTime must be a datetime object or None'
     if eTime is None:
-        eTime = sTime+datetime.timedelta(days=1)
+        eTime = sTime + datetime.timedelta(days=1)
     assert(sTime < eTime), "eTime must be greater than sTime!"
     assert(coords == 'gate' or coords == 'rng' or coords == 'geo' or
            coords == 'mag'), "error, coords must be one of 'gate', 'rng', " \
@@ -317,19 +317,21 @@ def plot_rti(sTime, rad, eTime=None, bmnum=7, fileType='fitex',
         figtop = .77
         if ((eTime - sTime) <= datetime.timedelta(days=1)) and \
                 (eTime.day == sTime.day):
-            figheight = .72/len(params)
+            figheight = .72 / len(params)
         elif ((eTime - sTime) > datetime.timedelta(days=1)) or \
                 (eTime.day != sTime.day):
-            figheight = .70/len(params)
+            figheight = .70 / len(params)
 
         for p in range(len(params)):
-            # Use draw_axes to create and set formatting of the axes to plot to.
-            pos = [.1, figtop-figheight*(p+1)+.02, .76, figheight-.02]
+            # Use draw_axes to create and set formatting of the axes to
+            # plot to.
+            pos = [.1, figtop - figheight * (p + 1) + .02, .76,
+                   figheight - .02]
             ax = draw_axes(rti_fig, data_dict['times'][fplot], rad,
                            data_dict['cpid'][fplot], bmnum,
                            data_dict['nrang'][fplot],
                            data_dict['frang'][fplot], data_dict['rsep'][fplot],
-                           p == len(params)-1, yrng=yrng, coords=coords,
+                           p == len(params) - 1, yrng=yrng, coords=coords,
                            pos=pos, xtick_size=xtick_size,
                            ytick_size=ytick_size, xticks=xticks,
                            axvlines=axvlines)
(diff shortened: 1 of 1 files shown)


MuhammadVT:
Hi @asreimer ,

I confirm that the bug exists in the current develop branch and this pull request fixes the problem. But if you look at the CPID panel, there is an overlay of two adjacent CPID numbers. This may need to be addressed, though not very important.  

Here is another relevant issue I have noticed: If you run the following command, you will see a weird CPID panel in the plot. I think this is a bug, but would you help me confirm whether it is a bug or not. Thank you. 

from davitpy.pydarn.plotting import rti
import datetime as dt
stime = dt.datetime(2010, 1, 15)
etime = dt.datetime(2010, 1, 16)
params = ['velocity']
coords = 'gate'; scales=[[-120,120]]
rad = 'bks'
#rad = 'sas'
bmnum = 7
channel = None
file_type = 'fitex'
colors = 'aj'
gsct=True; low_gray=True
file_name = None
fig = rti.plot_rti(stime, rad, eTime=etime, channel=channel, params=params, colors=colors,
        gsct=gsct, low_gray=low_gray, bmnum=bmnum, coords=coords, fileName=file_name,
        fileType = file_type, scales=scales)

Here is the plot,
bug_cpid


By the way, I did not have a chance to test your pull request about the unit testing. I will test it within one or two days.

asreimer:
Hey @MuhammadVT,

The problem you found seems to be caused by another issue, so I suggest submitting another issue on it (keeps discussions on here more focused). Looks like CPID is being plotted every minute for some reason (should only be plotted if it changes) and it looks like there is something else plotted on top of the text (see attached figure where I've zoomed in on the CPID axis):

figure_1

salsrag:
When you get solid areas like see for CPID and freq, I suggest that you try running a shorter plot, perhaps 1 hour or less. The freq  plot is wide because the radar is most likely switching between 2 frequencies. This produces a square wave, which when squished together with the 24 hour time axis produces a broad line. The CPID may be related in some way.

 

Ray

ksterne:
Hey @asreimer,

I can confirm that the current develop branch fails on the test you've given.  The test does work using the pull request.  I am a bit confused though on why/where the code is crashing though.  Is the gap in elevation angle data at around 1:10 utc?  Just hard to tell where the problem is from looking at the result.

Its true there is an issue found here that is noted in #186, but that is not the focus of this pull request.

Also, just went through styling the rti.py file for PEP8.  Can both of you pull this commit down and retest to make sure I didn't break anything?  It seems to be working here for me.  If you confirm, I will merge it in.

asreimer:
The reason it was failing is due to [these lines of code](https://github.com/vtsuperdarn/davitpy/blob/bugfix_rti_none_data_error/davitpy/pydarn/plotting/rti.py#L1068-L1075). Notice that I've added a `pArr[i] is None` to the conditional on the first line.

If there is data, `pArr[i]` is a list, but if not, then it is a NoneType. At 1:10 UT, the radar switched into a mode without elevation angles and so the original code was trying to index `pArr[i]` with `pArr[i][j]`, but you can't index a NoneType, so it crashed.

ksterne:
My question was a little more high level, but I think you answered it.  We're plotting elevation angles here, and the problem is that there isn't data.  I guess it wasn't totally clear that at 1:10 utc the elevation angle data would go away.  Looking at it on a full day it's hard to see it, I only found it by zooming in. I like these kinds of catches though!

Still waiting for confirmation that the PEP8'ing didn't break something...

asreimer:
Looks like the elevation angle data is missing simply because there was a change of operating modes from 3503 to 3370 (with no xcf) to 3503 again. Krieger may know why this happened.

asreimer:
Also, confirming that your PEP8 edits don't seem to break anything when I tested the code.

ksterne:
Ok, I'm probably over complicating the issue.  As I explained to @MuhammadVT, the "error" you mention here is like saying when I try to plot data from next week, I get an error.  But yet, at first glance we have data from next week.  Does that make sense?  I guess I was confused that there was somewhere that didn't have elevation angle data.

Anyways, sounds good.  I'll merge it now.

PR #169, opened 2015-10-28
Improve Non-VT Database use

aburrell:
Added two parameters to rcsetup, which allows users to specify
different databases for datafiles and hardware files.

Summary of changes:
davitpyrc: 
    1) added keywords SDBREADUSER, SDBREADPASS
    2) changed default for AACGM_DAVITPY_DAT_PREFIX to be more intuitive
    3) PEP8 compliance

rcsetup:
    1) added keywords SDBREADUSER, SDBREADPASS
    2) PEP8 compliance

pydarn/radar/radInfoIo.py:
    1) Lines 254-257 changed from DBREADUSER/PASS to SDBREADUSER/PASS
    2) a lot of PEP8 compliance changes.

Testing:
- If you use VT, set up using defaults and ensure you can still download data and load hardware info
- If you don't use VT, change the DBREADUSER defaults to your own server and load data from there.  Then use VT to load the hardware info.

PR #189, opened 2015-12-02
rewrite readScan to handle patterned scans

### LINKED ISSUE #181: Can't get scans from radDataPtr, what am I doing wrong?
I can't figure out why this isn't working. I know there is supposed to be data from this time.

```
Python 2.7.10 |Anaconda 2.1.0 (64-bit)| (default, Oct 19 2015, 18:04:42) 
Type "copyright", "credits" or "license" for more information.

IPython 4.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
%guiref   -> A brief reference about the graphical user interface.

In [1]: import datetime as dt

In [2]: from davitpy import utils, pydarn
Loaded davitpyrc file from DaViTpy installation directory. Path: /home/christer/.local/lib/python2.7/site-packages/davitpy-0.4-py2.7-linux-x86_64.egg/davitpy
Radars information has been updated.

In [3]: dtStart = dt.datetime(2013, 11, 3, 19, 0)

In [4]: data = pydarn.sdio.radDataRead.radDataOpen(dtStart, 'han', eTime=dtStart+dt.timedelta(hours=1))

Looking locally for fitex files with radcode: han channel: None
could not find fitex data in local files

Looking locally for fitacf files with radcode: han channel: None
could not find fitacf data in local files

Looking locally for lmfit files with radcode: han channel: None
could not find lmfit data in local files

Looking on the remote SFTP server for fitex files
Checking file: /tmp/sd/20131103.1800.01.han.fitex
Checking file: /tmp/sd/20131103.1800.11.han.fitex
Checking file: /tmp/sd/20131103.2001.00.han.fitex
removing invalid file: /tmp/sd/20131103.1800.01.han.fitex
removing invalid file: /tmp/sd/20131103.2001.00.han.fitex
found fitex data on sftp server
Concatenating all the files in to one
cat /tmp/sd/20131103.1800.11.han.fitex > /tmp/sd/20131103.190000.20131103.200000.han.fitex
rm /tmp/sd/20131103.1800.11.han.fitex

In [5]: scan = pydarn.sdio.radDataRead.radDataReadScan(data)

reached end of data

In [6]: type(scan)
Out[6]: NoneType
```

### LINKED ISSUE #182: Feature request: Pattern based scan contruction
Sometimes SuperDARN radars are run in a patterned type scan mode as can be seen with the Hankasalmi radar on `datetime(2013, 11, 3, 19, 0)` (cpid:153, see https://github.com/vtsuperdarn/davitpy/issues/181). 

Since it is still desireable to be able to plot FOV plots of these radar modes, perhaps one could make a `readScanPattern()` function that would take a pattern and return a scanData object for it. For example, if the pattern of beam numbers was: 15, 9, 14, 9, 13, 9, 12, 9, 11, 9, 10, 9, 9, 9, 8, 9, ... then with the appropriate arguments, `readScanPattern()` would return a scan consisting of beams: 15, 14, 13, 12, 11, 10, 9, ...

### LINKED ISSUE #183: Handle scan flag = -1
As @ksterne mentioned [here](https://github.com/vtsuperdarn/davitpy/issues/181#issuecomment-156207825):

>The scan flag being -1 is used by several other radars than just han. That I know of here is sye, sys, > unw, zho. I wonder if other Leicester radars do the same.... We may want to actually take a look at this.

With the current implementation of the `readScan` function in radDataTypes.py file (specifically [here](https://github.com/vtsuperdarn/davitpy/blob/develop/davitpy/pydarn/sdio/radDataTypes.py#L510-L550)), radars with the standard scan pattern but with -1 for the scan flag instead of 1 will not return a valid scan. This is fixed by adding an `abs()` in the `readScan` logic so that it doesn't care if the flag is 1 or -1.

So the questions is, is this a bug? Is this something we should be dealing with in davitpy or is this a data management problem? 

My biggest question is why some of the radars have non-standard scan flags?

cmeeren:
fixes #181
fixes #182
fixes #183

Changes:
* complete rewrite of readScan
* detects new scan based on non-zero scan flag AND same
  beam number as start of previous scan (see [code comment](https://github.com/vtsuperdarn/davitpy/blob/76be9a18571aaa278951d48f87ce847ea3d8d895/davitpy/pydarn/sdio/radDataTypes.py#L586-L593))
* manual scan pattern possible by specifying first beam
  (firstBeam) and every N beams (useEvery)
* automatic scan pattern detection by looping through some
  values of firstBeam and useEvery and checking for increasing
  or decreasing beam numbers in the resulting subset of beams
* raises ValueError if automatic detection failed
* tested and works for some consecutive scans from pyk and han
  at `datetime(2013, 11, 3, 19, 0)`

### DIFF
--- a/davitpy/pydarn/sdio/radDataTypes.py
+++ b/davitpy/pydarn/sdio/radDataTypes.py
@@ -507,47 +507,124 @@ class radDataPtr():
       from davitpy.pydarn.dmapio import setDmapOffset 
       return setDmapOffset(self.__fd,0)
 
-  def readScan(self):
-        """A function to read a full scan of data from a :class:`pydarn.sdio.radDataTypes.radDataPtr` object
+  def readScan(self, firstBeam=None, useEvery=None, showBeams=False):
+    """A function to read a full scan of data from a :class:`pydarn.sdio.radDataTypes.radDataPtr` object
 
-        .. note
-        This will ignore any bmnum request.  Also, if no channel was specified in radDataOpen, it will only read channel 'a'
+    .. note::
+      This will ignore any bmnum request in :func:`~pydarn.sdio.radDataRead.radDataOpen`.
+      Also, if no channel was specified, it will only read channel 'a'.
 
-        **Returns**:
-        * **myScan** (:class:`pydarn.sdio.radDataTypes.scanData`): an object filled with the data we are after.  *will return None when finished reading*
+    Parameters
+    ----------
+        firstBeam : int, optional
+            If manually specifying a scan pattern, will start picking beams at this index in the scan
+        useEvery : int, optional
+            If manually specifying a scan pattern, will pick every `useEvery` beam
+        showBeams : bool, optional
+            `showBeams` will print the collected scan numbers. Useful for debugging or if you manually want to
+            find the correct combination of `firstBeam` and `useEvery`.
 
-        """
-        from davitpy.pydarn.sdio import scanData
-        from davitpy import pydarn
-        #Save the radDataPtr's bmnum setting temporarily and set it to None
-        orig_beam=self.bmnum
-        self.bmnum=None
+    Returns
+    -------
+        myScan : :class:`~pydarn.sdio.radDataTypes.scanData` or None
+            A sequence of beams (``None`` when no more data are available)
 
-        if self.__ptr is None:
-            print 'Error, self.__ptr is None.  There is probably no data available for your selected time.'
+    Notes
+    -----
+
+    For patterned scans (e.g. if beam numbers are [5, 0, 5, 1, 5, 2, ...]) the function will try to find
+    a subset of the beams where beam numbers are increasing/decreasing by 1 throughout the scan.
+    Alternatively you can specify the pattern manually by using `firstBeam` and `useEvery`. You will
+    then get the subset of the beams starting at `firstBeam` (which is the beam's index in the list
+    of beams in the scan, not the beam number) and only including every `useEvery` beam.
+
+    """
+
+    from davitpy.pydarn.sdio import scanData
+    from davitpy import pydarn
+
+    #Save the radDataPtr's bmnum setting temporarily and set it to None
+    orig_beam = self.bmnum
+    self.bmnum = None
+
+    if self.__ptr is None:
+        print 'Error, self.__ptr is None.  There is probably no data available for your selected time.'
+        self.bmnum = orig_beam
+        return None
+
+    if self.__ptr.closed:
+        print 'error, your file pointer is closed'
+        self.bmnum = orig_beam
+        return None
+
+    myScan = scanData()
+
+    # get first beam in the scan
+    myBeam = self.readRec()
+    if myBeam is None:  # no more data
+        self.bmnum = orig_beam
+        return None
+    while not myBeam.prm.scan:  # continue to read until we encounter a set scan flag
+        myBeam = self.readRec()
+        if myBeam is None:  # no more data
+            self.bmnum = orig_beam
             return None
 
-        if self.__ptr.closed:
-            print 'error, your file pointer is closed'
-            return None
+    # myBeam is now the first beam we encountered where scan flag is set
+    myScan.append(myBeam)
+    firstBeam = myBeam.bmnum
 
-        myScan = scanData()
-        while(1):
-            myBeam=self.readRec()
-            if myBeam is None: 
-                break
+    # get the rest of the beams in the scan
+    while True:
 
-            if ((myBeam.prm.scan == 1 and len(myScan) == 0)         # Append a beam if it is the first in a scan AND nothing has been added to the myScan object. 
-             or (myBeam.prm.scan == 0 and  len(myScan) > 0) ):      # Append a beam if it is not the first in a scan AND the myScan object has items.
-                myScan.append(myBeam)
-                offset = pydarn.dmapio.getDmapOffset(self.__fd)
-            elif myBeam.prm.scan == 1 and len(myScan) > 0:          # Break out of the loop if we are on to the next scan and rewind the pointer to the previous beam.
-                s = pydarn.dmapio.setDmapOffset(self.__fd,offset)
-                break 
+        # get current offset (in case we have to revert) and next beam
+        offset = pydarn.dmapio.getDmapOffset(self.__fd)
+        myBeam = self.readRec()
+        if myBeam is None:  # no more data
+            break
 
-        if len(myScan) == 0: myScan = None
-        self.bmnum=orig_beam
-        return myScan
+        # Scan detection algorithm: We have a new scan if scan flag
+        # is set AND beam number is the same as the first beam number
+        # in the previous scan.
+        # The latter condition is important since we can encounter
+        # patterned scans with
+        #   scan flags:    [1, 1, 0, 0, 0, 0, ...]
+        #   beam numbers:  [5, 0, 5, 1, 5, 2, ...]
+        # and we don't want to break out on the 2nd beam in this case.
+        if myBeam.prm.scan and myBeam.bmnum == firstBeam:  # if start of (next) scan
+            # revert offset to start of scan and break out of loop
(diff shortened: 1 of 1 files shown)


cmeeren:
Anyone want to take a look at this?

asreimer:
Hey @cmeeren,

I'm just preparing for AGU and haven't had much time for much else. I'll had a look at this and the code looks good, but I haven't yet tested the code (but I will soon).

asreimer:
Oh, forgot to ask about the docstring changes. The style of the docstring in this pull request is different than what we have in the rest of the codebase. Seems @aburrell has made a few docstring style changes similar yours in #169. What's up?

cmeeren:
We decided to go with numpydoc style a while ago, didn't we? I think @aburrell originally suggested this.

asreimer:
Ya sorry, I forgot about that. It was a long time ago haha :P 

I suggested to @aburrell that we could change over all the docstrings before the next major release. The idea is that we would freeze new pull requests to develop and merge in the ones that are there. Next we would work on bugfixes and release necessary edits. And then we could implement the docstring change in one fell swoop. Similarly, we could replace all the print statements with the respective logging function calls.

asreimer:
I tested this on `han` for `datetime(2013, 11, 3, 19, 0)` and it works.
Tested on `cve` for `datetime(2013, 11, 3, 19, 0)` and it works (returning a scan with 20 beams in it as expected).
Tested this on a `cly` run where the radar was camped on onlly 1 beam and the code returned an exception as expected.
Tested on 'sas' and it worked as expected too!

One request before this is merged, I would really like the code to say something to the user if a "non-standard" scan pattern is found in the data (ie. one where the detected `useEvery` is not 1). That way, the user is made aware that there is s funny thing going on in the data and have fair warning about what is going on. 

Is it worth letting the user know when the scanflg is -1? I'm still baffled as to why an apparently Boolean thing (scanflg is **supposed** to be a flag, ie. True or False) is -1...

asreimer:
By the way, cool logic on detecting the scan pattern.

Why range only up to 10 in the `itertools.product` line of code though?

cmeeren:
**Scan flag -1:** With the (lack of) knowledge we currently have, I see no reason to bug the user if the scan flag is -1. It's trivial to implement this later if there turns out to be a good reason for it.

**Notifying the user if auto-detected `firstBeam != 0 or useEvery != 1`:** If the parameter `showBeams` is `True`, the auto-detected (as well as manually specified) scan pattern will be printed (`firstBeam`, `useEvery` and a list of the collected beam numbers). Is this sufficient? I am strongly in favor of giving the user control over the verbosity (a pet peeve of mine: using progress bars when looping is impossible if each iterations prints to the console). Alternatively we might add another parameter `warnNonStandard=True` which will show a warning if a non-standard scan pattern is found (e.g. using `warnings.warn()` - if so, this specific warning will only be shown once each session. Alternatively we can just use `print`.)

I only use `range(10)` because the highest values I have seen (I haven't looked very well) are `useEvery=2`, `firstBeam=1`, and I thought there is probably no patterned scans where the "real" consecutive scan to be extracted uses fewer than every 9 beams or starts after beam index 9. If you have reason to think there might be, I'm happy to increase the numbers.

**Possible bug, please check and comment**: In the second part of the code (scan pattern detection) the input variable `firstBeam` is used. However, the input variable seems to be overwritten in the first part of the code (scan fetching) to hold the beam number (not index) of the first beam in the scan. This looks like a bug to me; I can't see why I should have done that on purpose. Anyone else with me on this?

asreimer:
The more I look at this pull request, the slicker it seems. Python is awesome.

I like your `warnNonStandard=True` idea. Instead of using `warnings`, I think we should implement it as a print statement and then we can change it to use the appropriate `logging` level when we implement that while working on the 0.5 release over Christmas. 

Just to be safe (prevent some bug from some special mode in the future), and since it's super fast anyway, maybe a range(24) should be used [since some radars have 24 beams in them](http://vt.superdarn.org/tiki-index.php?page=Radar+Overview). Is that acceptable?

cmeeren:
Updated. Also fixed the bug I mentioned. Please test again and see if everything works like before.

asreimer:
@cmeeren those fixes you did are great.

Some further results from testing this:
Trying to construct a custom scan from sas data using firstBeam=3, useEvery=1. I expect to get a scan that looks like `[3,4,5,6,7,8,9,10,11,12,13,14,15,0,1,2]` but I get `[3,4,5,6,7,8,9,10,11,12,13,14,15]`. Here's the code I used:

    myPtr = pydarn.sdio.radDataOpen(datetime(2013,11,3,19),'sas',eTime=datetime(2013,11,3,20))
    myScan = myPtr.readScan(firstBeam=3,useEvery=1)
    print [x.bmnum for x in myScan]

One last thing, should it be that firstBeam and useEvery must be both specified in a manual mode? I found that I can specify just one and readScan happily ignores it and spits out its own thing.

cmeeren:
@asreimer the scan pattern you got is correct. I do not intend for this to "shift" the scan as you expected it to do. It still ends the scan when it encounters the new scan flag.

To be precise: First, an entire scan is collected. An "entire scan" is defined as everything between to beams with the same beam numbers and non-zero scan flag. THEN these beams are filtered according to `firstBeam` and `useEvery`.

If you want to manually specify a scan, you should use both. I do not see the point in having defaults of e.g. 1 and 2 here, because the whole point of this is to handle non-standard scans. If the user wishes to specify the pattern manually, I see no problem in having to specify both (in fact I think it's good, because it requries the user to be more aware of what they are doing).

It is of course possible to add a warning when only one of them is specified, saying that it will be discarded and the scan auto-detected.

asreimer:
Ok, now I understand the intended behaviour. I noticed this is described in the Notes in the docstring, but I've added an initial statement of this intended usage to the beginning of docstring of the function. I'm not so sharp here at AGU it seems haha.

I don't agree with the point about it simply being "trivial" for the user to construct a custom scan. By extension, that applies to this whole pull request and I think this pull request is a useful thing. It is similarly "trivial" to extract a custom scan when there's this funny scan pattern in our data. :)

No issue with both having to be specified, my point on the `firstBeam` and `useEvery` keywords was that they don't do anything if only one of them is specified and the docstring doesn't explicitly say then both need to be. I've added this in the docstring and added an assert to clear this up.

If the changes I made are a problem I can reverse them!

asreimer:
I should add that testing of the added assert code works for me. If it works for you @cmeeren I can merge this.

cmeeren:
@asreimer assertions are not the best way to check inputs. Raising a `ValueError` is more semantic. (I also simplified the condition). Furthermore, in Python one should almost never check types. Python uses "duck typing" - if it looks like a duck and quacks like a duck, then for all intents and purposes, it's a duck. In other words, if the user wishes to use `1` and `0` instead of `True` and `False`, that is their choice.

cmeeren:
What I meant by trivial is that this PR is more than 100 lines of code, whereas shifting the scan (which by the way would require rethinking the entire approach in this function) requires only these three lines (shifting by two beams):

```
beams_scan1 = myData.readScan()
beams_scan2 = myData.readScan()
shifted_scan = beams_scan1[2:] + beams_scan2[:2]
```

asreimer:
Haha duck typing. Never heard of that, I'll have to read up. If we don't want to use asserts anymore then we should have a discussion about this and make our codebase consistent.

I'll check your changes after lunch, but just for fun to read the patterned scan from that funny control program that motivated this pull request, one could have simply done:

    myScan = myPtr.readScan()
    myExtractedScan = [myScan[x] for x in range(len(myScan)] if x%2==0] 

or something like that to get every other scan. With python almost anything is trivial.

cmeeren:
Sure, but that assumes one knows the correct pattern. The new and juicy stuff in this function is the auto-detection of scan pattern, which required some more lines of code. :) But I agree though, Python makes a lot of things really compact and nice.

asreimer:
Everything looks good! I'm merging this in now.

Issue #184, opened 2015-11-12
Installing master branch (0.4) from previous version, package issue


ksterne:
So, I was transitioning our main davitpy code for the VT lab and I ran into an issue where upon running the `sudo python setup.py install` , I got the following error:

error: package directory 'gme' does not exist

as it was looking for a ~/davitpy/gme/ directory instead of ~/davitpy/davitpy/gme/.  It seems as though somewhere something got locked so that it wasn't transitioning to the new file structure, even though I blew away the repository and started from new (even deleting the /usr/local/lib/python2.7/dist-packages/davitpy* directories).  It seems as though the part of the code filling up the sources list wasn't adding davitpy to the beginning of the packages as it should, for example: davitpy.models, davitpy.models.iri .

So, two ways around this for this error.  The first is specific to this problem and would break everything else.  After the sources list I added:

    sources = [ 'davitpy.' + s for s in sources ]

which added the davitpy to everything and seems to get things working.  

The other one here, which is maybe more of an enhancement and could work for everyone.  Someone looked at that part of the code and wondered if it was even necessary.  If you import find_packages from setuptools and then change the line in the big setup call to:

    packages=find_packages(),

then populating the sources list isn't necessary.  I think the only caveat is that the find_packages() only works if there is an `__init__.py` file is in each package that you want to install.  Does anyone else have any experience with this and why we would want to use the nested for loops to populate a list of packages that we want to sent to setup?  Maybe this is something we can add to the next release.

asreimer:
Interesting. I think it is __very__ weird that suddenly now there are problems with installing davitpy even though nothing has changed in how the sources files are being added (since the last master+setup release). The weirdness factor is increased when you consider that none of us ever ran in to any install issues like this (that were reported at least) while developing this code in the develop branch... @ksterne is this an issue you run in to for just the current master branch or also the develop branch?

What do you mean by:
>as it was looking for a ~/davitpy/gme/ directory instead of ~/davitpy/davitpy/gme/.
what was looking for `~/davitpy/gme/`? and how do you know it was? Is this on Ubuntu or a different OS?

One thing to test is this:
1) Navigate to the directory in which you have the davitpy repository (for me this is /home/ashtonsethreimer/davitpy for example)
2) Fire up a python session and run the following:

    import os
    pwd = os.getcwd()
    sources = []
    source_dirs = ['davitpy']
    for s in source_dirs:
        for root, dirs, files in os.walk(pwd+'/'+s):
[...]

PR #190, opened 2015-12-03
Enhancement setup find packages

ksterne:
This addresses a very small case where the davitpy code is installed on the root directory as noted in #184.  Makes use of the find_packages() function from setuptools which is already being used in the setup.py code.  Also PEP8 styled the setup.py code.

Going with this may be more preferred than changing the current loops-to-find code.  Though I've this code in there but commented out.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import glob
 # Need to use the enhanced version of distutils packaged with
 # numpy so that we can compile fortran extensions
 from setuptools.command import install as _install
+from setuptools import find_packages
 from numpy.distutils.core import Extension, setup
 from numpy.distutils import exec_command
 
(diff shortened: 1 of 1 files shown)


asreimer:
Testing this pull request on my machines (Ubuntu 14.04, SUSE 13.2 and Fedora 23) and I have no issues installing and executing davitpy.

Tried installing davitpy from "/davitpy" (which used to fail as described in #184) also works just fine thanks to this pull request.

Since this is such a minor change I think we are ok to merge this right now.

**Something to important to note is that this pull request changes the davitpy version number to 0.5. You that means you will have 2 different davitpy packages installed if you don't remove the old one after installing this one.**

PR #198, opened 2015-12-21
0.5-release

asreimer:
This is the pull request for the 0.5-release.

This release includes:
1) RTI plotting improvements and bugfixes
2) Upgrade to HWM14 from HWM07
3) FOV categorization code (@aburrel's method)
4) More PEP8 compliance edits

Still remaining to be done (the only commits allowed to this branch):
1) Change docstring style to numpy format (i.e. http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_numpy.html)
2) Implement logging instead of print statments

Happy coding!

PR #199, opened 2016-01-07
document string, pep8, and logging changes

aburrell:
Changed document strings to numpy style
Made some pep8 changes (spaces, extra parenthesis)
Added logging and removed verbose keyword


cmeeren:
It seems you  forgot to remove the `verbose` keyword from where the functions are called, e.g. in radDataTypes.py. Mind fixing?

ksterne:
Took care of this over in the 0.5-release (#198) pull request.  Looks like there were a few in uncompress_files as well here.  There are more in some of the other `/sdio/` files.

PR #228, opened 2016-03-14
Release 0.4 master backup (into master-old)

ksterne:
I'll get the party started by making this backup request here.  I think I've done it right, but maybe @asreimer can confirm and finish things off.


asreimer:
Yes! Thanks for doing this @ksterne

PR #234, opened 2016-04-11
Freetype dependency in matplotlib

paalge:
Matplotlib depends on freetype. This includes a install of freetype6-dev.

**Note by @asreimer**:
Matplotlib requires freetype in order to compile properly. This dependency was missing in the `debian_dependencies.sh` script.

PR #235, opened 2016-04-12
Freetype dependency in matplotlib (#234)

asreimer:
Matplotlib requires freetype in order to compile properly. This dependency was missing in the `debian_dependencies.sh` script.


asreimer:
Hotfix that was merged into `master`, merging this into `develop` too.

Issue #214, opened 2016-02-10
sdDataTypes not using rcParams


aburrell:
As it says, sdDataTypes is hard-wired for certain options, such as tmpDir on line 151.

ksterne:
I guess without anything else happening on the other issues, I'll try to take this one on.

asreimer:
Fixed by #242

PR #242, opened 2016-04-22
Bugfix sd data types rc params

ksterne:
As @aburrell noticed in #214, the tmpdir wasn't being set by the rcParams value.  This has been corrected as well as added documentation to the sdDataPtr class.  Also noted which variables have been set at non-optional based on how the code is setup and/or assert error messages.

I don't have a test for this, although I guess you could set the DAVIT_TMPDIR in your davitpyrc file to something other than /tmp/sd/ .  Then run this code in the develop branch and see that using this file doesn't use the directory set in davitpyrc.

To see that this branch doesn't break anything, at least you could do:

```
from davitpy import *
import datetime

sillyObj = pydarn.sdio.sdDataOpen(datetime.datetime(2015,11,11), hemi='north', fileType='grdex')
```

And then see that the grdex files are in the tmpdir of your choosing.

### DIFF
--- a/davitpy/pydarn/sdio/sdDataTypes.py
+++ b/davitpy/pydarn/sdio/sdDataTypes.py
@@ -42,7 +42,7 @@ class sdDataPtr():
 
     Parameters
     ----------
-    sTime : Optional[datetime]
+    sTime : datetime
         start time of the request
     hemi : str
         hemisphere of data interested in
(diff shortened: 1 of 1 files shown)


asreimer:
Confirmed that this fixes the issue. Again, not a lot of code was changed so I'm just going to merge this.

Issue #244, opened 2016-05-02
[Critical Bug] Broken Exception Handling


asreimer:
Since we made the changes over to the logging module in the last release, there are now problems with our exception handling. Specifically, if there is an exception one gets another exception, so you don't get to see what the original exception problem (hooray!). This is what has been happening in #241 and #238, I was just too dense earlier to realize it.

"TypeError: coercing to Unicode: need string or buffer, exceptions.TypeError found"
and
"cannot concatenate 'str' and 'exceptions.ImportError' objects"

We need a hotfix for this ASAP. I'm going to work on it now.

asreimer:
The second error is easily fixed by converting the exception to a string. For example:

    try:
        print h
    except Exception, e:
        logging.warn('PROBLEM: ' + e)
Fails, but this

    try:
        print h
    except Exception, e:
        logging.warn('PROBLEM: ' + str(e))
Works. I'm converting all of these types of statements to include the explicit str() conversion.

ksterne:
Merged it into master and created pull back to develop to resync with #247.

Issue #241, opened 2016-04-21
TypeError when do mapDatObj.overlayCnvCntrs()


Shirling-VT:
Hi all,

I found an error when running the convection plotting code in our SuperDARN Data Plotting notebook. It seems that there's some type error somewhere when we try to overlay convection contours in a map plot. The following shows that error message. Anyone can confirm this and has any idea on how to fix this? It seems exist in both present master and develop branch.

---

TypeError                                 Traceback (most recent call last)
<ipython-input-4-70e57d4c587e> in <module>()
     11 mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax)
     12 mapDatObj.overlayMapFitVel()
---> 13 mapDatObj.overlayCnvCntrs()
     14 mapDatObj.overlayHMB()

/usr/local/lib/python2.7/site-packages/davitpy-0.6-py2.7-macosx-10.11-x86_64.egg/davitpy/pydarn/plotting/plotMapGrd.pyc in overlayCnvCntrs(self)
    656             vmax=potCntr.max(), vmin=potCntr.min(),
    657             colors = 'DarkSlateGray', linewidths=1.,
--> 658             locator=LinearLocator(12) )
    659         plt.clabel(cntrPlt, inline=1, fontsize=10)
    660         return cntrPlt

/usr/local/lib/python2.7/site-packages/mpl_toolkits/basemap/**init**.pyc in with_transform(self, x, y, data, _args, *_kwargs)
    528             # convert lat/lon coords to map projection coords.
    529             x, y = self(x,y)
--> 530         return plotfunc(self,x,y,data,_args,_*kwargs)
    531     return with_transform
    532 

/usr/local/lib/python2.7/site-packages/mpl_toolkits/basemap/**init**.pyc in contour(self, x, y, data, _args, *_kwargs)
   3633                 mask = np.logical_or(ma.getmaskarray(data),xymask)
   3634                 data = ma.masked_array(data,mask=mask)
-> 3635                 CS = ax.contour(x,y,data,_args,_*kwargs)
   3636         except:
   3637             ax.hold(b)

/usr/local/lib/python2.7/site-packages/matplotlib/**init**.pyc in inner(ax, _args, *_kwargs)
   1810                     warnings.warn(msg % (label_namer, func.**name**),
   1811                                   RuntimeWarning, stacklevel=2)
-> 1812             return func(ax, _args, *_kwargs)
[...]

Issue #238, opened 2016-04-19
TypeError on input


heinet:
When I try to input davitpy, I get the following error.  This occurs even when I try to import the GOES notebook.

Populating the interactive namespace from numpy and matplotlib
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-7aca422ea469> in <module>()
      3 from matplotlib import pyplot as plt
      4 import datetime
----> 5 from davitpy import gme #This is the DavitPy GeoMagnetic Environment module

/Users/heinet/Library/Python/2.7/lib/python/site-packages/davitpy-0.5-py2.7-macosx-10.6-x86_64.egg/davitpy/__init__.pyc in <module>()
    616     from davitpy import pydarn
    617 except Exception, e:
--> 618     logging.exception('problem importing pydarn: ' + e)
    619 
    620 try:

TypeError: coercing to Unicode: need string or buffer, exceptions.TypeError found

In [2]:

spacecataz:
Can confirm on pure macports install.

asreimer:
I cannot reproduce this problem on the OpenSUSE nor Fedora.

I'm guessing there is some problem with your mac installation of davitpy. Likely, something is wrong in the pydarn module judging from the error message you have included: "618 logging.exception('problem importing pydarn: ' + e)"

In a python shell, what do you get if you run:
`from davitpy import pydarn` ?

Shirling-VT:
Hi all,

I just installed davitpy in my macbook though the mac_brew dependencies,
though almost manually after failing on several packages installation. But
it seems to work well here without the issue heinet mentioned. I also
suspect something was wrong during the installation. You have to be very
carefully when doing the dependencies installation and make sure every
required packages are installed successfully before you do the real
installation.

On Tue, Apr 19, 2016 at 2:28 PM, Ashton Reimer <notifications@github.com>
wrote:

> I cannot reproduce this problem on the OpenSUSE nor Fedora.
>
> I'm guessing there is some problem with your mac installation of davitpy.
> Likely, something is wrong in the pydarn module judging from the error
> message you have included: "618 logging.exception('problem importing
> pydarn: ' + e)"
>
> In a python shell, what do you get if you run:
> from davitpy import pydarn ?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/vtsuperdarn/davitpy/issues/238#issuecomment-212057791>
>



[...]

PR #246, opened 2016-05-02
[hotfix] broken exception handling

asreimer:
As detailed in #244 exception handling statements throughout the codebase were broken when logging was implemented in the last release. This needs to be fixed ASAP.

Also, this pull request fixes some of the problems that were encountered in #238 and #241.

After pulling this in to master, master will need to be pulled into develop.


ksterne:
Can confirm it seems like you've got them all.  I've been seeing this kind of error in some random cases.

PR #251, opened 2016-05-16
re-syncing develop from overlayFov hotfix

ksterne:
It seems as though the fixes for the exception strings as well as the debian dependencies is going back with this is well?  @asreimer, is this from where you branched off of develop before these changes were applied?

Otherwise, I think I set this up right...maybe I got it backwards.


asreimer:
Hey @ksterne, you did get this correct. That one mod to the debian_dependencies was also a hotfix.

PR #252, opened 2016-05-20
elevation and height error

aburrell:
Added elevation angle and virtual height error calculations to field-of-view updating routines.


aburrell:
Separated the elevation angle and virtual height routines into their own files, to ease use by other routines in the future.

Updated comments in 2 of the radar files to comply with current standards.

Changed the way tdiff is handled in the proc/fov/update_backscatter routines.

aburrell:
Merging

ksterne:
Why was this merged without any testing or comment?

aburrell:
I tested it and since it didn’t have any changes to non-comments in the parts of the code I hadn’t written figured that was fine.  If you think it needs more, feel free to revert.

asreimer:
I think it would be good to test this code on other machines before merging (if not for anything but continuing to follow our development model). I also think it's a good idea to provide example usage code and explain expected results of using the code to help others test it. I haven't had the time as of late for davitpy coding or I would have done some testing.

asreimer:
Reverting is messy and I don't think we should rely on simply pushing things and reverting if others disagree. My internet connection for the next week and a half isn't stable enough to do more than comment here, so I'll defer the reverting to someone else.

aburrell:
Fair points, but it does show a weakness in our workflow.  Namely only one of us typically has time for anything at any given time.  I think it would be a good idea to have everyone sign up for periods of time when they’ll be the default person to start testing code (though anyone and everyone can jump in) instead of everyone saying “not me” or overburdening one or two people (which seems to be the current method).  This is one of the reasons I have a load of analysis code that isn’t committed (but should be) and I know I’m not the only one.


As for testing this, I should have pointed out that any of the tests developed for fov (all the routines in test_update_backscatter) would be appropriate for this.  Sorry for not including that.

asreimer:
One other solution is to get more developers (maybe not realistic).

Perhaps even another solution is as I suggested about 6 months ago: segregate the codebase into 2 different codebases. 1 for code SuperDARN data reading/plotting code/utilities, and 2 for analysis tools, including the models. For the core reading and plotting utilities, we could have different rules for development than for the analysis tools. For example, really, you and @w2naf are the only one's who understand what the MUSIC and extended FOV plotting codes do/should do. That makes those chunks of the code harder for anyone else to do development on right from the get go. Here it might make sense to be a little looser the development rules for MUSIC and the FOV stuff, specifically if the main author of those codes has approved changes. This would only be clean if they were in their own repo though.

It's time for another davitpy developer meeting...

asreimer:
@aburrell I think you need to remake your pull request after the revert.

aburrell:
Yeah, at this point it’s too much faff.  I’ll just make this a part of a larger change I’m going to push out (hopefully) during the superdarn meeting.  I’d prefer to push small bits of themed changes, but this one is massive enough that I might as well not.

ksterne:
Hey @aburrell,

Definitely understand the mild frustration of slow development cycles with our code.  Part of this is the brain drain we've been experiencing with others going off and working in industry or on other projects.  However, I think to keep things as stable as possible, we should continue to follow our development model.

I try to get to things that are easy and are more of an ask to run on different machines.  In the past, things that I thought were very simple and minor took several weeks for someone to get around to it.  I took care of that with a weekly "bump" of commenting here to get it back to the top of everyone's inboxes.  I suspect that doing this a couple of times can help draw attention to it.  Otherwise, if it's a busy week and flooded with e-mails, it gets lost.

I get the concern that there are parts of davitpy that are more specific to certain research and maybe not useful for much of anyone outside of one person.  But I'm wondering how much duplication we'll need to have if they are split into two parts.  I don't think this specific analysis code is hindering our ability to develop the code.  More so, it's less time and the pull request not presented such that someone is able to run the code quickly, easily, and provide feedback to the developer.  This is where I see a good need for unit tests, or example test code.

With the development model we're using, I don't see it as a call for others to completely understand what the new code is doing.  More so, others are asked to test it out since we're using different configurations.  This is a way to keep code from only being compatible with one operating system or one version of python.

I'm certainly good with another developer meeting, though for next week I won't be there until early, early Tuesday.

asreimer:
I agree. It can be frustrating at times trying to get pull requests merged, both due to the speed and due to the bugs/hidden problems others find. It would be really great if we could get a few more developers working with us (been trying here and successfully had our software engineer build a python based dmap reading/writing library). Mostly, we just need polish to get davitpy to a 1.0. The convection map plotting needs work, and the unit testing pull request just needs more unit tests added to it. Is anyone at the SuperDARN meeting going to talk davitpy?

A final note on my suggestion, splitting up the code doesn't require any duplication. There would be davitpy and davitpy-extras or something like that, with the latter depending on the former (like what is done for matplotlib and basemap).

I'm unavailable for a meeting until after June 5th. Perhaps we should organize it on the google group.

Issue #102, opened 2014-09-19
MLT plotting labelled with degrees, not MLT


mrwessel:
Maps plotted in MLT end up showing degrees from magnetic local midnight, not actual MLT.

ksterne:
I think we're good to close this old, old bug?

asreimer:
I don't know. I just tried making a convection map using my old testing code in #103 and the mlt labels are wrong.

aburrell:
Same here.  I think this is a problem that was never addressed.

asreimer:
While I don't remember explicitly and haven't looked through the commit history to verify, I merged #103 because it did fix the labeling issue. Something else maybe broke it?

Either way, it appears this MLT labeling issue still persists...

mrwessel:
Testing with python plotUtils.py produces the desired result for me.  Others please verify.

asreimer:
This is fixed in #102.

PR #260, opened 2016-06-12
Bugfix for #102:  Label MLT with hours, not degrees

mrwessel:
Fixed in #103 originally, but broken since then.

See #102.

One-liner fix that re-added the format kwarg to a function call.


asreimer:
Is this the proverbial fiery rise from the ashes for @mrwessel? Nice to see you contributing here again :)

I tested this code using `python plotUtils.py` as instructed and it works. MLT labels are no longer in degrees, but in MLT hours as they should be.

I also test the code using:

    pydarn.plotting.plotFan(datetime(2013,1,1,12),['sas'],coords='mlt')

and this works as expected as well.

Issue #233, opened 2016-04-06
SuperDARN Data Plotting Notebook and Underlying Functions


w2naf:
Hi, guys.

Nice job on the new release! I'm just starting to use the new code now.

I found that some of the changes break things in the SuperDARN Data Plotting Notebook (https://github.com/vtsuperdarn/davitpy/blob/35c3de481ff4a12699401b390d7d4e387592ead9/docs/notebook/SuperDARN%20Data%20Plotting.ipynb). I tried fixing some of these problems myself last night, but ran into some issues I want to get feedback on before changing anything.

The main problem is that the figure keyword was removed from plotting.rti.plot_rti() (https://github.com/vtsuperdarn/davitpy/blob/35c3de481ff4a12699401b390d7d4e387592ead9/davitpy/pydarn/plotting/rti.py#L87). It looks like this was done because the new code loops through a frequency band filter and is then supposed to return a list of figures.
- I personally advocate against this, as I think these plotting routines should just return a single object. That sort of looping should be done in a user-level script or a wrapper.
- If we do want to keep this functionality, then there are some issues in plot_rti() we need to fix.

Next, the overlay of potential contours is now broken. I think this is because of bug in matplotlib.ticker.LinearLocator(), and filed and issue regarding that here: https://github.com/matplotlib/matplotlib/issues/6270

Thanks,
Nathaniel

Shirling-VT:
Hi Nathaniel,

I agree with you about the rti plot issue. I also find it inconvenient
without the figure keyword when I tried to write a script to get more than
one rti plot and save the figures for later usage. The lack of figure
keyword prevents me from using of savefig function.

Best wishes,
Xueling

On Wed, Apr 6, 2016 at 10:41 AM, Nathaniel Frissell <
notifications@github.com> wrote:

> Hi, guys.
>
> Nice job on the new release! I'm just starting to use the new code now.
>
> I found that some of the changes break things in the SuperDARN Data
> Plotting Notebook (
> https://github.com/vtsuperdarn/davitpy/blob/35c3de481ff4a12699401b390d7d4e387592ead9/docs/notebook/SuperDARN%20Data%20Plotting.ipynb).
> I tried fixing some of these problems myself last night, but ran into some
> issues I want to get feedback on before changing anything.
>
> The main problem is that the figure keyword was removed from
> plotting.rti.plot_rti() (
[...]

PR #262, opened 2016-06-21
fix the SuperDARN_Data_plotting.ipynb

MuhammadVT:
This pull request fixes SuperDARN Data Plotting.ipynb

To text this pull request,  run SuperDARN Data Plotting.ipynb with the current develop branch and after installing this pull request. 

The convection plot in the last part of the notebook may not work as mention in #233 due to the bug in https://github.com/matplotlib/matplotlib/issues/6270, which is reported by @w2naf . The is fixed in 
https://github.com/matplotlib/matplotlib/pull/6425, however, the bugfix  has not been merged into the current matplotlib master branch. If you want to get the convection mat plot work you may do the following:

git clone git@github.com:matplotlib/matplotlib.git
cd matplotlib
git fetch
git checkout -b v1.5.x origin/v1.5.x
sudo pip install -e .

Just in case if the figure does not show up in the ipython console, then you may check ....../matplotlib/lib/matplotlib/mpl-data and change  backend      : %(backend)s into
backend      : Qt4Agg


MuhammadVT:
Just a reminder, before you run "sudo pip install -e ." you need to "cd matplotlib"

asreimer:
This pull request does not modify `plot_rti` to address the "returning of multiple figure objects when multiple frequency bands are supplied" part of issue #233. That will still have to be fixed in a subsequent pull request.

asreimer:
Hi @MuhammadVT 

Is the matplotlib pull needed just to fix contour plotting in the convection maps? Just trying to clarify what I should expect in testing this.

MuhammadVT:
Hi @asreimer ,

Yes, you need to have the v1.5.x branch of matplotlib installed to fix contour plotting.

ksterne:
Good to see this notebook getting worked on!  I was just showing it to someone the other day....or at least trying to.  Are we going to have to make more changes to this notebook once #233 is fixed?

Also, should we be making this fix until things are fixed in with https://github.com/matplotlib/matplotlib/pull/6425?   For now it's a bit awkward that if someone asks us why something doesn't work that we have to have them do something that should be included in the debian_dependencies.sh script and/or accepted in the matplotlib repo.  This only relates to the very last line in the notebook as @asreimer is noting.

I think it's worth a wait on the matplotlib pull request.

MuhammadVT:
New release of matplotlib with the bugfix included has been released. 
https://github.com/matplotlib/matplotlib/releases

@ksterne, would you mind testing this notebook with the new version of matplotlib installed.

ksterne:
@MuhammadVT,

Are the issues with matplotlib fixed with #245 ?

asreimer:
Just a note that `pip install --upgrade matplotlib` currently only installs matplotlib 1.5.1, NOT 1.5.2. despite 1.5.2 being officially shipped on July 3rd. I've asked about if/when 1.5.2 will be available in pip.

So, to answer your question @ksterne, no, the current dependency scripts WILL NOT resolve the issue (which is annoying...).

asreimer:
Tested this pull request. First this is what I did to get matplotlib 1.5.2 installed (release created on github, but they haven't included in pip yet):

    git clone https://github.com/matplotlib/matplotlib.git
    cd matplotlib
    git checkout v1.5.2
    sudo python setup.py install

This is how I recommend we install matplotlib 1.5.2. It could be added the scripts in #245 to replace the pip installation until such a time as 1.5.2 is available in pip.

Not sure why @MuhammadVT used `pip install -e`. Could you explain?

Otherwise, this pull request resolves the issue, although I'm not sure what was changed in the data analysis notebook since the "bugfix" is to update matplotlib.

asreimer:
Let's wait a few days though, I think 1.5.2 will be in pip in the next day or two.

asreimer:
Alright, matplotlib 1.5.2 is now installable via pip. This means that rerunning the dependency installation script will resolve this bug and anyone installing davitpy from now on won't encounter this problem.

Testing this pull request I find that it resolves the issues that we had with the Data Plotting notebook.

Merging.

Issue #256, opened 2016-06-06
Bug in ae.py


GPerry306:
Keywords 'dbName' and 'collName' in /davitpy/gme/ind/ae.py should be 'dbname' and 'collname', respectively, according to 'getDbConn' function in /davitpy/pydarn/sdio/dbUtils.py.

GPerry306:
The GM indices notebook doesn't seem to work at all for me by the way.

MuhammadVT:
Hi @GPerry306 

Thank you for pointing out the bugs. I have created a pull request #257 to fix this issue. Could you please test it out.

Thanks,

Muhammad

asreimer:
Problem fixed in #257 which has been merged into `develop` now.

PR #265, opened 2016-07-09
Bugfix music and geomagind

w2naf:
This pull request fixes bugs:

- Variables were not passed as strings to the logger in the geomagnetic indices modules, causing them to break.

- musicRTI was not updated to the new DavitPy RTI API, causing MUSIC notebooks to break.

- datetimeToEpoch did not recognize np.ndarrays as interables

This pull request will make the MUSIC and MUSIC-simulate notebooks work properly again.


MuhammadVT:
Hi @w2naf,

Running MUSIC.ipynb breaks when the following commands are executed:
_# We can use the image processing library in scikit-image to automatically find the peaks representing signals.
pydarn.proc.music.detectSignals(dataObj)
pydarn.plotting.musicPlot.plotKarr(dataObj)

ImportError: cannot import name find_contours_


Also, some of the changes you made in geomag indices related files have been fixed in #256

ksterne:
@MuhammadVT, have you run ./debian_dependenicies.sh lately?  I originally got an error at the same point, although a different error than you.  However, I went back and re-ran the dependencies script and I'm getting passed that point.  I do get a bunch of VisibleDeprciationWarning's though that look like:

/usr/local/lib/python2.7/dist-packages/davitpy-0.6-py2.7-linux-x86_64.egg/davitpy/pydarn/proc/music/music.py:2124: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future

I also get some of these for the last cell's code `pydarn.plotting.plotKarrDetected(dataObj)`

Otherwise, this notebook does completely run for me.

MuhammadVT:
as @ksterne has pointed out, it happened to be a local problem with older version of skimage. deleting it from dist-packages and pip install scikit-image fixes the problem. Running MUSIC, MUSIC-simulate and plotGMagIndices.ipynb notebooks all worked fine. Merging it....

PR #282, opened 2016-10-26
Fixed the scatter part of fan

paalge:
Plotting was not done on an axis, meaning that the plotting crashed, as Figure does not have the method scatter.


ksterne:
Hello @paalge, a few notes here on this pull request:

- Have you had a chance to review our development structure?  If not, it's in https://github.com/vtsuperdarn/davitpy/tree/master/docs/development_instructions .  It's preferable that this type of pull request is made into develop instead of master as you've done here.

- In addition, can you describe more the problem and why this pull request should be merged in?  This could/should include code you use to demonstrate the problem (makes it easy for someone to test) and a screenshot of the figure highlighting where the problem occurs.

PR #284, opened 2016-11-02
back porting hotfix #245 into develop

ksterne:
Hopefully my git-fu muscles are getting stronger here.  Yet another reason we should keep on a regular schedule of making releases twice a year!

Here's the back porting of the installation hotfix issue as we saw from #245.  Does someone feel inclined to merge this pull request?  (Granted that I haven't messed something up)


asreimer:
Since this is just a branch to update develop from a hotfix that was patched to master, I'm merging it.

Issue #286, opened 2016-11-10
MLT conversion issue when doing the convection map


Shirling-VT:
Hi,

Here is the code when we do the convection map:

import datetime
import matplotlib.pyplot as plt
import davitpy.pydarn.plotting.plotMapGrd
from davitpy.utils import *
fig = plt.figure()
ax = fig.add_subplot(111)
sdate = datetime.datetime(2016, 3, 1, 9, 0)
mObj = plotUtils.mapObj(boundinglat=50., datetime=sdate,
                        gridLabels=True, coords='mlt')
mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax)
mapDatObj.overlayMapFitVel()
mapDatObj.overlayCnvCntrs()
mapDatObj.overlayHMB()
plt.show()

Usually, we set the mlt coordnates in the mapObj function, but we keep the parameter plotCoords in the davitpy.pydarn.plotting.plotMapGrd.MapConv rountine, with which we can choose between the mlt and mag coordinates as well, even though we have already set the coords in the mapObj. I found this really confusing and there are typos and other importing issues in the below code, which will give you an error message when you use plotCoords to set the coordinates:

        # check if hemi and coords keywords are correct
        assert(hemi == "north" or hemi == "south"), \
            logging.error("hemi should either be 'north' or 'south'")
        assert(plotCoords == 'mag' or **coords** == 'mlt'), \
            logging.error("error, coords must be one of 'mag' or 'mlt'")

        # mlt conversion stuff
        if self.plotCoords == 'mlt':
            epoch = timeUtils.datetimeToEpoch(**strtTime**)
            mltDef = aacgm.mltFromEpoch(epoch,0.0) * 15.
            lonShftFit += mltDef
            gridArr[1,:] = numpy.mod( ( gridArr[1,:] + lonShftFit ) / 15., 24. )
        else:
            gridArr[1,:] = ( gridArr[1,:] + lonShftFit ) 

[...]

PR #287, opened 2016-11-10
Xueling updated the python_install_mac_port.sh and plotMapGrd.py.

Shirling-VT:
The mac dependency installation script has been updated, though just tested in two macbooks. Some typo and importing issues are fixed in the plotMapGrd.py file as well, though the plotCoord parameter still does not work in the mlt coordinates.


ksterne:
Adding 'wontfix' label to this for now as all new pull request are frozen until the 0.6 release becomes official.

asreimer:
Confirming that this pull request fixes problem with the plotMapGrd. I also found another bug that is related to what @Shirling-VT discussed in #286. That is a separate issue from this pull request so I'll file a new issue and separate pull request for it.

Merging.

PR #290, opened 2016-12-08
Release 0.5 backup to master-old

ksterne:
Getting things started by making a "backup" for the 0.5 release here.  Mostly doing this for legacy, it's probably a general question if we should keep doing this.  The pro I can see is we can point someone to this release if they need something in a pinch.  The con is that we make release point tar/zip files anyways...


ksterne:
Anyone paying attention and wanting to do a merge here?  Otherwise, I'll self merge in 15 minutes.

PR #292, opened 2016-12-15
Enhancement acf plot input check

ksterne:
This is a small update for maybe something larger that we should strive for with davitpy routines.  Here I've added a first cut at input santiziation/checks for two of the `acfPlot.py` routines.  I've also added a few small PEP8 edits from doing a check on this file.  Must have missed these earlier.

To test this run:
```
from davitpy import pydarn
from datetime import datetime
myPtr = pydarn.sdio.radDataOpen(datetime(2012,5,21), \
                                      'kap',fileType='fitacf')
myBeam = myPtr.readRec()
pydarn.plotting.acfPlot.plot_acf(myBeam,'24')
```
and
```
from davitpy import pydarn
from datetime import datetime
myPtr = pydarn.sdio.radDataOpen(datetime(2012,5,21), \
                                         'kap',fileType='fitacf')
myBeam = myPtr.readRec()
pydarn.plotting.acfPlot.plot_rli(myBeam)
```
which have been shamelessly stolen from the examples in each routine (with a few minor tweaks to break things).  I guess conversely, you could run the unit test of `python acfPlot.py` and see that I haven't broken anything.  With the example though, you'll see the routines crash at some random point in the develop branch.  Whereas with this enhancement, you'll see you get some error messages from the assert lines.

### DIFF
--- a/davitpy/pydarn/plotting/acfPlot.py
+++ b/davitpy/pydarn/plotting/acfPlot.py
@@ -95,8 +95,8 @@ def plot_acf(myBeam, gate, normalized=True, mark_blanked=True,
 
     # Input checks
     # myBeam check for rawacf file
-    assert(myBeam.fType == 'rawacf'), logging.error('myBeam must '
-        'be from a rawacf file')
+    assert(myBeam.fType == 'rawacf'), logging.error(
+        'myBeam must be from a rawacf file')
     # Check of gate parameter
     assert(isinstance(gate, int) and gate >= 0), logging.error(
         'gate must be an integer and zero or positive')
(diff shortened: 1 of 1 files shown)


asreimer:
This is excellent @ksterne. I also totally agree that we should be doing proper input checking in all of our routines. Users will thank us later and it will also save us time in the long run because people won't be able to "break" the code using weird inputs.

I'm testing this just now.

asreimer:
Alright, I've tested this and I don't see any problems. It's nice that it's throwing errors as expected!

Since this pull request isn't super big, I'm going to merge it.

Issue #294, opened 2017-01-16
Installation of stable basemap


asreimer:
Currently, our dependency script is not installing any particular major release of basemap.

The offending line of code is:
`git clone https://github.com/matplotlib/basemap.git`

which should be:
`git clone --branch v1.0.7rel https://github.com/matplotlib/basemap.git`

This needs to be fixed because if you happen to clone basemap after they've made a change that breaks it, you install a broken basemap...

PR #295, opened 2017-01-16
Use full release of basemap

asreimer:
This fixes issue #294.

We have been installing a rolling release of basemap instead of a release. This hotfix fixes that.

When this pull request is finished, it will have to be pulled into develop also.


asreimer:
I have finished adding all the commits that this hotfix needs.

This doesn't need much testing. Just check to make sure that I didn't make any typos (ie. that the git clone command that I changed works).

Shirling-VT:
Confirmed and merged.

PR #296, opened 2017-01-17
Removed unnecessary plotCoords variable that caused bug.

asreimer:
As discussed to in #286, the plotCoords in `plotMapGrd.py` is actually unnecessary. Not only that but it actually causes a big problem in the convection map plotting when setting `plotCoords=mlt`. For example, running this code in `develop`:

    import datetime
    import davitpy.pydarn.plotting.plotMapGrd
    from davitpy.utils import *
    fig = plt.figure()
    ax = fig.add_subplot(111)
    sdate = datetime.datetime(2016, 3, 1, 9, 0)
    mObj = plotUtils.mapObj(boundinglat=50., datetime=sdate,
                            gridLabels=True, coords='mlt')
    mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax,plotCoords='mlt')
    mapDatObj.overlayMapFitVel()
    mapDatObj.overlayCnvCntrs()
    mapDatObj.overlayHMB()
    plt.show()

will produce a groce (with a c) convection map plot:
figure_1

After squashing this bug with this pull request, you should be able to make beautiful plots that look like this:

figure_2

The `plotCoords` keyword has been removed since it doesn't actually do anything anymore, so when you run your code on this pull request make sure you don't use it.

### DIFF
--- a/davitpy/pydarn/plotting/plotMapGrd.py
+++ b/davitpy/pydarn/plotting/plotMapGrd.py
@@ -34,8 +34,6 @@ class MapConv(object):
     maxVelScale : Optional[float]
         maximum velocity to be used for plotting, min is zero so scale is
         [0,1000]
-    plotCoords : Optional[str]
-        coordinates of the plot, only use either 'mag' or 'mlt'
 
     Attributes
     ----------
@@ -52,7 +50,7 @@ class MapConv(object):
         the axis handle used
     mObj : utils.plotUtils.mapObj
         the map object you want data to be overlayed on.
- 
+
 
     Methods
     --------
--- a/davitpy/pydarn/plotting/plotMapGrd.py
+++ b/davitpy/pydarn/plotting/plotMapGrd.py
@@ -122,12 +122,12 @@ class MapConv(object):
         # requested
         if hemi == "north":
             assert(mObj.boundarylats[0] > 0.), \
-                logging.error("Map object is using one hemisphere and data the"
-                              " other")
+                logging.error("Map and data objects must be from the same"
+                              " hemisphere")
         else:
             assert(mObj.boundarylats[0] < 0.), \
-                logging.error("Map object is using one hemisphere and data the"
-                              " other")
+                logging.error("Map and data objects must be from the same"
+                              " hemisphere")
 
         # check if hemi and coords keywords are correct
         assert(hemi == "north" or hemi == "south"), \
--- a/davitpy/pydarn/plotting/plotMapGrd.py
+++ b/davitpy/pydarn/plotting/plotMapGrd.py
@@ -103,7 +103,7 @@ class MapConv(object):
 
 
     def __init__(self, startTime, mObj, axisHandle, hemi='north',
-                 maxVelScale=1000., plotCoords='mag'):
+                 maxVelScale=1000.):
 
         import datetime
         from davitpy.pydarn.sdio import sdDataOpen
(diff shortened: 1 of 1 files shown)


Shirling-VT:
Confirmed the bug in the develop branch and if run the following codes in the bugfix_plotCoords_variable branch:

import datetime
import davitpy.pydarn.plotting.plotMapGrd
from davitpy.utils import *
fig = plt.figure()
ax = fig.add_subplot(111)
sdate = datetime.datetime(2016, 3, 1, 9, 0)
mObj = plotUtils.mapObj(boundinglat=50., datetime=sdate,
                                          gridLabels=True, coords='mag')
mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax)
mapDatObj.overlayMapFitVel()
mapDatObj.overlayCnvCntrs()
mapDatObj.overlayHMB()
plt.show()

image


import datetime
import davitpy.pydarn.plotting.plotMapGrd
from davitpy.utils import *
fig = plt.figure()
ax = fig.add_subplot(111)
sdate = datetime.datetime(2016, 3, 1, 9, 0)
mObj = plotUtils.mapObj(boundinglat=50., datetime=sdate,
                                          gridLabels=True, coords='mlt')
mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax)
mapDatObj.overlayMapFitVel()
mapDatObj.overlayCnvCntrs()
mapDatObj.overlayHMB()
plt.show()

image


import datetime
import davitpy.pydarn.plotting.plotMapGrd
from davitpy.utils import *
fig = plt.figure()
ax = fig.add_subplot(111)
sdate = datetime.datetime(2016, 3, 1, 9, 0)
mObj = plotUtils.mapObj(boundinglat=50., datetime=sdate,
                                          gridLabels=True, coords='mlt')
mapDatObj = davitpy.pydarn.plotting.plotMapGrd.MapConv(sdate, mObj, ax)
mapDatObj.overlayGridVel()
mapDatObj.overlayCnvCntrs()
mapDatObj.overlayHMB()
plt.show()

image

asreimer:
Excellent.

Since @Shirling-VT and I worked on this together, it's probably a good idea for someone else to test this code and merge it.

aburrell:
Looking through the plotMapGrd.py file there are some changes that need to be made before merging:

1) MapConv description no longer match input (reference to plotCoords variable still present)
2) Error statement, "Map object is using one hemisphere and data the other" could be changed to be more clear, perhaps: "Map and data objects must be from the same hemisphere"
3) overlayGridVel, calcFitCnvVel descriptions are incomplete
4) nitpick: are we really going to default to jet?  https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/
5) Parenthesis should be added on 199 (vecLen = velsPlot[nn] * self.lenFactor / self.radEarth / 1000.)
6) Note for future: PEP8 changes still needed here.

Ran the tests and was able to reproduce all three plots. 

dtest1
dtest2
dtest3

ksterne:
Wow, it's been awhile!  Sorry my code development time has been taken up by the demands happening in RST world.  I've added two commits here to address @aburrell's 1 and 2.  

For 3, I see this is outside of the scope of this bugfix.  Yes it should be corrected, but doesn't relate here.  

For 4, I think this would take some kind of central SuperDARN authority on what default color schemes would be....good luck on finding or being able to organize that.  I think the idea has been that slow motions are cool colors while hot colors represent fast motion.

For 5, @aburrell, agreed, even though this is outside of the scope, it may be worth it.  Do you feel comfortable placing the parentheses in the correct spots?

Otherwise, I've tested this and see that it's working.  So aside from the last comment here, this is ready to merge.

ksterne:
I should also note that the latest dependencies update throws a lot of depreciation warnings from matplotlib.  So, we'll be looking at a good overhaul of code to clear those out.

ksterne:
Post it here and I can push it up?

aburrell:
Not practical.  Better to just leave it for a different pull request, since parens won’t affect the code functionality.

ksterne:
Sounds good. Merging it now.  I'll try to get to the aacgm pull request before next Thursday's release branch creation.

PR #310, opened 2017-06-01
Post 0.7 release master back into develop

ksterne:
I don't think there was any real commits aside from the merge conflict resolution that should be done here.  Though I see there's a couple of files relating to the mac installation of basemap.  Anyone care to clarify this and merge it in if this is good?


ksterne:
Anyone care to make this happen or should I self merge here?

Shirling-VT:
Just merged.

Issue #321, opened 2017-09-09
overlayFov is broken by changes to fov generation code


asreimer:
Currently, in the `develop` branch, overlayFov is unable to plot FOVs for all SuperDARN radars without producing wildly wonky results. 

![superdarnfovs](https://user-images.githubusercontent.com/4604819/30244294-5173c626-9570-11e7-8009-54523ab2e70f.png)

Here's the code I used to produce this figure:

    from davitpy.pydarn.radar import *
    from davitpy.pydarn.plotting import *
    from davitpy.utils import *
    import datetime as dt
    from matplotlib import pyplot
    
    fig = pyplot.figure(figsize=(16,8))
    ax1 = fig.add_subplot(121)
    m1 = plotUtils.mapObj(boundinglat=30., gridLabels=True,ax=ax1)
    overlayRadar(m1, fontSize=8, plot_all=True, markerSize=5)
    overlayFov(m1,plot_all=True)
    ax2 = fig.add_subplot(122)
    m2 = plotUtils.mapObj(boundinglat=-30., gridLabels=True,ax=ax2)
    overlayRadar(m2, fontSize=8, plot_all=True, markerSize=5)
    overlayFov(m2,plot_all=True)
    fig.show()

If you turn on the logging to be `debug-annoying` you will see a TON of warnings like this:

`WARNING:root:Accuracy on height calculation (5.0) not reached quick enough. Returning nan, nan.`

This is an issue that goes back until 0.5-master. 0.5-master is able to make sane looking plots. Everything after that has this bug that makes the crazy plots. Something changed in the fov generation code between then and now that is producing NaNs when it really shouldn't be.

asreimer:
Since this affects 2 of our master branch releases, I'd say this is a critical bug that needs a hotfix.

aburrell:
Agreed.  I'm happy to do some work on it once I get the two pull requests I'm working on done.

aburrell:
Ok, I found where the bug is and know how to fix it.  Pull request coming soon...

aburrell:
Addressed in #324

mts299:
I assume this issue is fixed now?

aburrell:
THanks for the reminder

PR #312, opened 2017-06-21
Python 2/3 compatibility in Python code, streamline install doc, implement continuous integration, vastly simplify install

scivision:
* C code still uses obsolete calls to Python lib, that is blocker for Python 3. 
* removed some syntax errors in code

PR #324, opened 2017-10-10
FoV plotting bug hotfix

aburrell:
Field point calculation at large distances was breaking FoV plotting, since the more distant locations weren’t meeting the conditions for reliability.  I fixed this by adding a flag that defaults to skipping
this evaluation.  I would recommend setting this flag to True when performing studies where one actually cares about the location, but this level of accuracy is not what we're looking for when making FoV plots.

### DIFF
--- a/davitpy/pydarn/radar/radFov.py
+++ b/davitpy/pydarn/radar/radFov.py
@@ -474,7 +474,7 @@ class fov(object):
 def calcFieldPnt(tr_glat, tr_glon, tr_alt, boresight, beam_off, slant_range,
                  adjusted_sr=True, elevation=None, altitude=None, hop=None,
                  model=None, coords='geo', gs_loc="G", max_vh=400.0,
-                 fov_dir='front'):
+                 fov_dir='front', eval_loc=False):
     """Calculate coordinates of field point given the radar coordinates and
     boresight, the pointing direction deviation from boresight and elevation
     angle, and the field point slant range and altitude. Either the elevation
@@ -522,8 +522,13 @@ def calcFieldPnt(tr_glat, tr_glon, tr_alt, boresight, beam_off, slant_range,
         location 'I' for groundscatter (default='G')
     max_vh : (float)
         Maximum height for longer slant ranges in Standard model (default=400)
-    fov_dir
+    fov_dir : (str)
         'front' (default) or 'back'.  Specifies fov direction
+    eval_loc : (bool)
+        Evaluate the calcualted location based on reasonable tolerances (True)
+        or accept the first calculation (False).  Using True gives better
+        locations, but restricts data at the furthest range gates.
+        (default=False)
 
     Returns
     ---------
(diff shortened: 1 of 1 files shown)


aburrell:
TEST based on issue #321 

import davitpy
import datetime as dt
plt.ion()
stime = dt.datetime(2016,6,1)
f = plt.figure(figsize=(16,8))

ax1 = f.add_subplot(121)
m1 = davitpy.utils.plotUtils.mapObj(boundinglat=30.0, gridLabels=True, ax=ax1)
davitpy.pydarn.plotting.overlayRadar(m1, fontSize=8, plot_all=True, markerSize=5)
davitpy.pydarn.plotting.overlayFov(m1, plot_all=True)

ax2 = f.add_subplot(122)
m2 = davitpy.utils.plotUtils.mapObj(boundinglat=-30.0, gridLabels=True, ax=ax2, datetime=stime)
davitpy.pydarn.plotting.overlayRadar(m2, fontSize=8, plot_all=True, markerSize=5)
davitpy.pydarn.plotting.overlayFov(m2, plot_all=True)

fov_bug_map

scivision:
As a note for future modernization, there is a [Python standard library syntax for declaring type hints](https://www.python.org/dev/peps/pep-0484).
That is, to declare the type of a variable.

    def func(a:int, b:float) -> float:

[MyPy](http://mypy-lang.org/) is a powerful linter using these type hints.
[PyCharm IDE](https://www.jetbrains.com/help/pycharm/type-hinting-in-pycharm.html) uses type hints.

In the future, JIT compilers may use these type hints.

asreimer:
It works! But I don't want to merge until #312 is merged.

aburrell:
Since #312 is being reformated, I think we should merge this in.  Thoughts, @asreimer ?

aburrell:
@asreimer , @ksterne shall we merge this in?  Not great to have bugs lying around.

ksterne:
Sorry for the delay here @aburrell!  Lots of other things taking up my time.  Maybe I can work on this this week.

scivision:
Yes merge is my opinion. I will make my (unrelated) changes in smaller pull requests

ksterne:
Hey @aburrell, I'll give my :( for different remote switching.  But I know you usually operate out of your own repo.
pr324_after_pgrrkn
While it looks better, there's still some issues with plotting pgr and rkn here.  There might be one other one, but I can't see it quickly.  Do you know off hand what's causing that or is it easy to look into?  If it's outside the scope of disabling the fine accuracy for locating scatter, then I'm seeing the same as you are.
pr324_after

aburrell:
From what I remember, those are the radars that have a maxgate of 225.  So, if you don't specify a maximum range gate of 75 or 100 or so, they take over a lot of the globe.  If you limit their range gates, they should look sensible.

ksterne:
Ah, right! I always forget those values that were put in and makes things a little strange.  One day it might be neat to change the maxGate parameter to a list that follows the radar codes/IDs.  But that's another day for sure.  Anyways, using the maxGate parameter everything looks sensible for me!  @asreimer?  Or anyone else?

aburrell:
Ok, looking back at the comments now that I've had several cups of tea, it looks like @asreimer and @ksterne have tested.  Would one of you like to merge?

ksterne:
I can get to it, but not until tomorrow.  If @asreimer does before me then no problem.

ksterne:
OK, getting to merge this now.  @aburrell, do you want to make the PR to pull this to develop as well?

aburrell:
Added on #344 .

Issue #309, opened 2017-06-01
install issue with davitpy_path


arfogg:
After (seemingly) successfully running install there is an error importing davitpy that stems from the rcPARAM davitpy_path.  @aburrell has commented this out in the past to get around this issue, but that's not a real solution for everyone.

    $ python setup.py install --user
    $ipython
    
    In [1]: import davitpy as dp
    ---------------------------------------------------------------------------
    RuntimeError                              Traceback (most recent call last)
    <ipython-input-1-cccfcdb2eb5a> in <module>()
    ----> 1 import davitpy as dp
    
    /home/a/arf18/source/davitpy/davitpy/__init__.py in <module>()
        608 
        609 # Set up the rcParams dictionary
    --> 610 rcParams = rc_params()
        611 
        612 #TO DO
    
    /home/a/arf18/source/davitpy/davitpy/__init__.py in rc_params(fail_on_error)
        504         return ret
        505 
    --> 506     return rc_params_from_file(fname, fail_on_error)
        507 
        508 

    /home/a/arf18/source/davitpy/davitpy/__init__.py in rc_params_from_file(fname, fail_on_error, use_default_template)
        592     iter_params = six.iteritems(defaultParams)
        593     config = RcParams([(key, default) for key, (default, _) in iter_params
    --> 594                                       if key not in _all_deprecated])
        595     config.update(config_from_file)
        596 
    
    /home/a/arf18/source/davitpy/davitpy/__init__.py in __init__(self, *args, **kwargs)
        377     def __init__(self, *args, **kwargs):
        378         for k, v in six.iteritems(dict(*args, **kwargs)):
    --> 379             self[k] = v
        380 
        381     def __setitem__(self, key, val):
    
    /home/a/arf18/source/davitpy/davitpy/__init__.py in __setitem__(self, key, val)
        391                 return
        392             try:
    --> 393                 cval = self.validate[key](val)
        394                 if key == 'verbosity':
        395                     level_dict = {'silent': 60,
    
    /home/a/arf18/source/davitpy/davitpy/rcsetup.pyc in validate_path_exists(s)
         85         return s
         86     else:
    ---> 87         raise RuntimeError('"%s" should be a path but it does not exist' % s)
         88 
         89 
    
    RuntimeError: "" should be a path but it does not exist
    
    In [2]: debug
    > /home/a/arf18/source/davitpy/davitpy/rcsetup.py(87)validate_path_exists()
         85         return s
         86     else:
[...]

PR #325, opened 2017-10-10
Removed problematic and unused rcParam

aburrell:
Removed the rcParam ‘DAVITPY_PATH’, which has caused installation
issues and is not used anywhere at all (that I could find).

This pull request was made in response to issue #309 .

### DIFF
--- a/davitpy/rcsetup.py
+++ b/davitpy/rcsetup.py
@@ -230,7 +230,6 @@ defaultParams = {
                                  validate_string],
     'IGRF_DAVITPY_COEFF_FILE':	[model_coeffs_dir + 'igrf/igrf12coeffs.txt',
                                  validate_string],
-    'DAVITPY_PATH':             [path, validate_path_exists],
 
     # the verbosity setting for logging
     #'verbose.level': ['silent', validate_verbose],


asreimer:
Let me take a look before this gets merged. I originally coded up the rcParam stuff so I can test it. I'll try to look this weekend.

aburrell:
Thanks Ashton!

asreimer:
I'm investigating, but I don't think we can simply remove this rcParam.

As a naive check, running `grep -lr "DAVITPY_PATH" *` I see that `DAVITPY_PATH` is actually used in a few places, specifically:

    davitpy/rcsetup.py
    davitpy/models/iri/__init__.py
    davitpy/models/hwm/__init__.py
    davitpy/models/hwm/hwm_input.py
    davitpy/models/raydarn/rt.py

This is because this path is set so that those models can find the coefficient files that they need to run.

The approach here seems like, DAVITPY_PATH is causing problems so let's just delete it, but we don't know why it's causing the problems. I would much rather see us get to the bottom of the problem so it can be properly addressed.

I'm going to do more investigating, but I'm also closing this pull request.

PR #326, opened 2017-10-16
Use absolute paths instead of relative

asreimer:
This pull request fixes the bug reported in #309.

The problem was caused because relative paths were used instead of absolute paths. See https://github.com/vtsuperdarn/davitpy/issues/309#issuecomment-336820639 for more details.

Testing is:
1) Without this pull request, try importing davitpy from within the davitpy source directory (if you have the source at /home/user/davitpy, then cd there and try to import). This will fail.
2) With this pull request, try 1). It will work.

Thanks to @arfogg for reporting.

### DIFF
--- a/davitpy/rcsetup.py
+++ b/davitpy/rcsetup.py
@@ -214,9 +214,10 @@ validate_verbose = ValidateInStrings(
     'verbose',
     ['silent', 'helpful', 'debug', 'debug-annoying'])
 
+# Installation path of davitpy (based on location of this file)
+path = os.path.abspath(os.path.split(os.path.dirname(__file__))[0])
 
 # determine install location of model coefficients
-path = os.path.split(os.path.dirname(__file__))[0]
 model_coeffs_dir = os.path.join(path, 'tables/')
 
 if not os.path.exists(model_coeffs_dir):


scivision:
This seems fine just from looking. In the future, use of Python standard library `pathlib` (actually `pathlib2` for py27 compatibility) will cleanup these long path code lines and remove ambiguity because `path` will be an object instead of just a string. Effectively zero time penalty for using `pathlib.Path` vs. `str`.

[Reference using pathlib vs str](https://www.scivision.co/python-idiomatic-pathlib-use/)

asreimer:
That is a nice writeup @scivision. Thanks for sharing!

aburrell:
Since we didn't want to merge this before because of the py3 pull request that is being re-done, I say we merge this in now.  Thoughts @asreimer ?

ksterne:
This code looks good and the test works out.  It's likely it will be updated, but until then, this would be good to get the fix in.  Can someone else give it a test and merge it it?

aburrell:
You're the third person to test it @ksterne , so I think it's ok to merge now.  Shall I do the honours?

ksterne:
Oh, I didn't know others had actually tested it.  I thought you'd just made a comment we could move forward since python3 is being separated out.  I didn't think there was a test done by @scivision, but again, maybe just didn't read it right.  Don't know why we would hold off here.

aburrell:
I was thinking of you, me, and @asreimer , but my counting is wrong since @asreimer as the creator cannot test :-P

ksterne:
Yea, self testing shouldn't quite count (as they should have it working before making the PR).  It wasn't clear that you were good with it @aburrell, but if you're good then go ahead and merge it.  This fix stems from a problem that I think you were associated or partially observed in #309.

aburrell:
Mmmm… yes, I’ll test again with better comments so that our protocol remains satisfied and if it’s successful, merge.  Sorry for my sloppy comments on this pull request.

ksterne:
Eh, not a problem, it's not like it's a lot of code to be changed. Just couldn't quite tell.

aburrell:
Alright, just tested again and it still works after the fix, but not before.  Merging.

PR #334, opened 2018-01-25
Enhancement acfPlot plotting updates

ksterne:
I know things are fairly quiet, but here's a fairly minor and simple update I came across when using the acfPlot code.  Basically just making the resulting plot look a little prettier and usable.  Things that I changed (in case you're too lazy to read the commit messages):

- Set x-axis tick marks for lag numbers to integers, because I've never heard of lag 2.5
- Added a legend for ACF plotting, makes it clearer which is which.
- Added zero-filling the gate number for pdf/png saving.  So instead of filename5.pdf it's filename05.pdf for a gate 5 plot.
- Added to title of plot the range gate which is good if you're looking at multiple gates or want to concatenate multiple plots in one document later on.

Sample test code:

```
from davitpy import pydarn
import datetime as dt

myPtr = pydarn.sdio.radDataOpen(dt.datetime(2017,9,17,1,20), 'fhe', eTime=dt.datetime(2017,9,17,1,40), fileType='rawacf', bmnum=11)

myBeam = myPtr.readRec()

pydarn.plotting.acfPlot.plot_acf(myBeam, 25, png=True, show=False)

```

You should get something like:

acf_20170917_012057_ut_fhe_gate25

Run the same code on the develop and see the differences between the two.  Not great, but this one presents the info/data a little better.

### DIFF
--- a/davitpy/pydarn/plotting/acfPlot.py
+++ b/davitpy/pydarn/plotting/acfPlot.py
@@ -193,10 +193,10 @@ def plot_acf(myBeam, gate, normalized=True, mark_blanked=True,
             if (png is False) and (pdf is False):
                 png = True
             fig = mpl_fig()
-        ax1 = fig.add_axes([0.1, 0.55, 0.35, 0.35])
-        ax2 = fig.add_axes([0.1, 0.1, 0.35, 0.35])
-        ax3 = fig.add_axes([0.5, 0.1, 0.35, 0.35])
-        ax4 = fig.add_axes([0.5, 0.55, 0.35, 0.35])
+        ax1 = fig.add_axes([0.12, 0.55, 0.35, 0.35])
+        ax2 = fig.add_axes([0.12, 0.1, 0.35, 0.35])
+        ax3 = fig.add_axes([0.52, 0.1, 0.35, 0.35])
+        ax4 = fig.add_axes([0.52, 0.55, 0.35, 0.35])
 
         rad_name = pydarn.radar.network().getRadarById(myBeam.stid).name
         if xcf:
--- a/davitpy/pydarn/plotting/acfPlot.py
+++ b/davitpy/pydarn/plotting/acfPlot.py
@@ -232,7 +232,7 @@ def plot_acf(myBeam, gate, normalized=True, mark_blanked=True,
         ax1.plot(lags, im, marker='o', color='green', lw=2, label='Imag')
         ax1.plot([lags[0], lags[-1] + 1], [0, 0], 'k--', lw=2)
 
-        ax1.legend(loc='lower right')
+        ax1.legend(loc='lower right', fontsize='medium', ncol=2)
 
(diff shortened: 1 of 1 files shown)


Shirling-VT:
Hi Kevin,

Here is the plot I get from your testing code:
image

Shirling-VT:
This is what I get from the develop branch:
image

You can see that the legend in the left first panel and the range gate info in the title are gone. The x axis of lag number is also different as suggested by Kevin's comments.

aburrell:
May I suggest two changes?  One is to have an additional zero pad for the range gate number on the files, since some radars have a maximum gate of 225.  The other is to decrease the font size on the legend (probably specify "medium" or "small") since it looks like it's larger than the axes labels.

ksterne:
Hey @aburrell, thanks for the feedback here.  Those are easy things to change up and I've done them now.  I've also forced the acf plot's legend to 1 one and 2 columns as I think that will look better on more plots by keeping it away from the data points.  What do you think?

aburrell:
This is better.  Here's what I get for running the code now:

acf_20170917_012057_ut_fhe_gate025

The last cosmetic changes I think would help are:

1) Make the title two lines, with a "\n" either after the radar name or after the date
2) Remove the comma after the month (AGU date standard is "17 Sep 2017")
3) Change from "m/s" to "m s$^{-1}$", again for plot publishing standards.
4) Top left y-axis is almost off the plot, decreasing the white space between the right and left columns would help.

ksterne:
OK @aburrell, I can take it that there's some basic things to fix up here, but starting to creep away from the original intent.  Things look good after the last couple of commits I pushed?  Only way to solve 4) was to move all of the plots to the right a little which makes it a little off balance, but still looks good to me.

aburrell:
This looks good to me, I think the output is usable now.

acf_20170917_012057_ut_fhe_gate025

ksterne:
So....is this ready to be merged? @aburrell?

aburrell:
I say yes.  Shall I do the honours?

aburrell:
@ksterne , I'll let you decide whether or not to do the branch clean up.

alpha721:
hello @ksterne . I came across your project while I was hunting the gsoc page and got really interested in your project. I am pretty comfortable with python and was wondering if I could be a contributor. I read the guidelines, but was a little apprehensive about mailing you directly, ( for the project "Develop open source software in python to download, analyze and visualize space science datasets").

alpha721:
I would be really thankful  if you could give a general idea, on a few things:
i) when would be the right time to mail you regarding the project
ii) what are the expectations from an applicant in general . Suggestions on anything you find is worth mentioning really.
I would be grateful to you if you could just give a general advice on how to get started with the project. 
Thank you

ksterne:
E-mailing me directly would be much more appropriate than commenting on a closed pull request which can be sent to all as an update.  Please look at our ideas list tomorrow as our application process will be changing.

PR #336, opened 2018-02-12
removed bad asserts from fetchUtils.py

### LINKED ISSUE #330: bad asserts in fetchUtils
The `outdir` asserts found [here](https://github.com/vtsuperdarn/davitpy/blob/master/davitpy/pydarn/sdio/fetchUtils.py#L197) and [here](https://github.com/vtsuperdarn/davitpy/blob/master/davitpy/pydarn/sdio/fetchUtils.py#L433) are bad. 

Specifically the `outdir[-1] == "/"` part is nasty.

Using `os.path.join` automatically handles all of the slash problems with joining paths together. So we should remove this from the asserts.

Additionally, line [652](https://github.com/vtsuperdarn/davitpy/blob/master/davitpy/pydarn/sdio/fetchUtils.py#L652) needs to be changed to:

    tf = os.path.join(outdir, rf)

anu0012:
Fixed #330

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ It is easiest to have either homebrew (http://brew.sh/) or MacPorts (http://www.
 
 Next, follow the instructions for Ubuntu, but for the dependencies script, choose one of the following:
 
-    sudo ./python_install_mac_brew.sh
-    sudo ./python_install_mac_port.sh
+    sudo ./install/python_install_mac_brew.sh
+    sudo ./install/python_install_mac_port.sh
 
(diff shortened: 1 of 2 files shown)


aburrell:
Code looks good.  @anu0012 , I'll test this once you provide a few lines of code that demonstrates how this works with a few types of output directory strings.

asreimer:
Hey @anu0012, thanks for doing this. I've changed the base to `develop` instead. We'll get this in soon.

aburrell:
The asserts no longer ping and the filenames created are appropriate.

PR #343, opened 2018-03-02
Removing need to resource .bashrc or open/close terminal

ksterne:
Updating the install instructions in the `README.md` file as we're no longer explicitly modifying bash environment variables during the installation.  Though it is true that in order for the hdw.dat data to run properly, you need to set some environment variables which has caused a lot of problems related to that.  Maybe there's a bugfix coming where we need to add those to the davitpyrc file...but that's outside of this PR.

Test is to run through a fresh install and don't close the terminal window at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,6 @@ We need to install some dependencies using a script that makes calls to apt-get 
 
     sudo ./install/debian_dependencies.sh
 
-Note that this script also modifies your ~/.bashrc file to set environment variables identifying the DaViTPy installation location, the SuperDARN database access information, and others.  Because of this, please source your ~/.bashrc file (or close and reopen a new terminal window) to refresh your environment variables.
-
 Next, do the actual davitpy install (from in the davitpy directory):
 
     sudo python setup.py install


aburrell:
This works, but it has the old behaviour for DAVIT_PATH (where it doesn't allow you to open davitpy in the davitpy directory).

ksterne:
Sounds good @aburrell! You're right, the import davitpy in davitpy directory won't work on this branch since this was split from develop before #326 was merged in.

MuhammadVT:
Hey @ksterne,

The change is in README.md file only. Do you still want me to test this out?

ksterne:
Hey @MuhammadVT, yes the only change is to the readme file, but the test is to see if the instructions for getting things going with the whole codebase are good in that readme file.  Basically, fresh Ubuntu virtual machine install and see if you need to refresh the environment variables at any point during the install (so outside of what @aburrell noted).

MuhammadVT:
Hey @ksterne,
I have tested it on a fresh Ubuntu 16.04 virtual machine. As expected, no need to refresh the environment variables, but I had to run the following two commands 
__sudo apt-add-repository universe_
_sudo apt-get update__
Before running,
_sudo ./install/debian_dependencies.sh_
In order to deal with the error message of

> Unable to locate package python-pip

ksterne:
Sounds good @MuhammadVT, updating dependency info seems like something outside of what we should do for this repo.  If you're satisfied, looks like you're the second tester...so care to merge?

PR #344, opened 2018-03-02
field point calculation

aburrell:
Field point calculation at large distances was breaking FoV plotting,
since the more distant locations weren’t meeting the conditions for
reliability.  I fixed this by adding a flag that defaults to skipping
this evaluation.


aburrell:
This is the same bug fix that was added to the master branch in #324.

ksterne:
OK, sorry it's taken me a bit, but since @aburrell tested my pull request, I figured I'd follow up here.  Since it doesn't seem like there's been any changes since #324, I think this is good to go.  I did pull the code again and ran it again to to confirm my results from the pull request into `master`.

aburrell:
Cheers everyone!  Hitting the button.

PR #352, opened 2018-03-25
typo in rt.py

satyarth42:
output is mistakenly written as uotput in rt.py


ksterne:
It seems as though your branch may have been based on the `master` branch, instead it's better to start with the `develop` branch and then make commits to create a pull request back into `develop`.  Can you rebase your branch and re-submit this pull request?

asreimer:
Thanks for reporting this. I'm going to just directly make the change. This pull request has a couple unrelated changes that we don't want to make here.

PR #364, opened 2018-06-14
Update version number to 0.9 for upcoming release.

asreimer:
Very simple and should be self explanatory :)

### DIFF
--- a/davitpy/__init__.py
+++ b/davitpy/__init__.py
@@ -33,7 +33,7 @@ from itertools import chain
 import logging
 from davitpy.rcsetup import defaultParams
 
-__version__ = str('0.5')
+__version__ = str('0.9')
 
 _error_details_fmt = 'line #%d\n\t"%s"\n\tin file "%s"'
 
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,7 @@ data_files.append(('davitpy/models/hwm',
 # Now execute the setup
 #############################################################################
 setup(name='davitpy',
-      version="0.8",
+      version="0.9",
       description="Space Science Toolkit",
       author="VT SuperDARN Lab and friends",
       author_email="ajribeiro86@gmail.com",


ksterne:
I've usually made this a commit in the release branch.  But I'm not sure what the best practice might be.

ksterne:
Ah, whoops!  I didn't look at what was actually being changed.  Seems as though there's two places where the version is set.  One in the `__init__.py` and the other is in `setup.py`.  I thought this was referencing the later.  Doh!  Would you be good with doing both in the release branch?

aburrell:
I agree with @ksterne that the version number (in both places) should be changed in the release branch.

asreimer:
I don't really understand the reasoning behind "only in the release branch". Isn't the develop branch the de facto next "release branch"?

Please educate me. Have we even discussed what the release branch procedure is? I've been out of the loop a bit the last year so I'm probably missing some key info. Feel free to move this discussion into private email.

ksterne:
The develop branch is working towards the next release, though I think it keeps things cleaner so that there's not confusing about what version you're using.  I like to think that the `develop` branch is built upon the previous release.  So right now the develop branch is improvements upon version 0.8 and at whatever point we decide to make a 0.9 release (we/I have been slacking on this), that's when you can update a version number.  Otherwise, if 0.9 is assigned to the develop branch, then you could stop using master and develop and use the version numbers instead.  Does that make some sense?

I don't think you've missed much in the last year or so on release branch procedures.  Only thing I think might have changed is we decided trying to do enhancements/improvements (like the documentation style change) in the release branch isn't a great idea.

ksterne:
In the interest of time, I'm changing what I thought before and lets go ahead here.  I've updated the version number in the `setup.py` file as well.  I'll give it a day or so in case someone catches an error I've made or didn't think about, but otherwise I'll merge this in.

ksterne:
Merging this in for prep for a final release for deprecation.

PR #365, opened 2018-06-14
[hotfix] update version number

asreimer:
The version number in the released master branch is wrong. This fixes it.

### DIFF
--- a/davitpy/__init__.py
+++ b/davitpy/__init__.py
@@ -33,7 +33,7 @@ from itertools import chain
 import logging
 from davitpy.rcsetup import defaultParams
 
-__version__ = str('0.5')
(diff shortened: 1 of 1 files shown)


ksterne:
Test here is in the `master` branch to run something like `python -c "import davitpy; print(davitpy.__version__)"` from bash command line or translate that into 2 ipython commands.  On `master` you'll get 0.5, with the hotfix branch you'll get the appropriate 0.8.  

As it might be apparent in #364, I wasn't even aware of this version number being set.  I had only paid attention to the other one as that is what dictates what directory structure the files are placed into.

I see this works here.  @asreimer, maybe pick someone for a second?  It's really a 5 minutes or less test.

PR #367, opened 2018-07-19
Fortran fixes

jhaiduce:
Hi,

A few fixes to Fortran code and wrappers here:

- Changes to geopack_08.pyf so that the field computed by dip_08 is returned to the calling python code
- Fixes to the MSIS and IRI code so that it compiles with GCC 8.1.1

I do not use MSIS or IRI and I wasn't able to find  test code in the davitpy repository, so I don't have a way to test those changes aside from verifying that the code compiles. Is there a test suite that I can run to verify that my changes didn't break anything?

John


ksterne:
Closing this for now as this repo has become deprecated.

